Issue #2, opened 2022-07-18
Unit tests with gdUnit (Godot 4.0 only)


bitbrain:
Currently, there is no way of telling if a change will break this addon or not. Introduce unit tests to ensure that any new PR will not break existing logic.

This should only apply to the `godot-4.x` branch since maintaining two separate versions of GUT is difficult to maintain and once Godot 4 launches, the `godot-3.x` branch goes into maintenance mode.

GUT is one option but after much consideration, gdUnit4 has a better and cleaner integration:

https://github.com/MikeSchulze/gdUnit4

## Tests to write

- [x] simple composites
- [x] advanced composites (star nodes)
- [x] advanced composites (random nodes)
- [x] decorators
- [x] complex behavior tree setup (end-2-end)
- [x] simple blackboard
- [x] advanced blackboard (shared between multiple trees)
- [x] interrupt logic for running nodes
- [x] beehave tree methods (currently running action + get last condition)

Nemesis-AS:
Which unit testing framework would you recommend?
Currently I'm aware of GUT - https://github.com/bitwes/Gut
and WAT - https://github.com/AlexDarigan/WAT
There may be other ones as well

bitbrain:
I ported GUT to Godot 4 so we should use that one!

https://github.com/bitwes/Gut/tree/godot_4

creadicted:
> I ported GUT to Godot 4 so we should use that one!
> 
> https://github.com/bitwes/Gut/tree/4.0

Do you mean https://github.com/bitwes/Gut/tree/godot_4 ?

bitbrain:
@creadicted good catch! It got renamed.

bitbrain:
I discovered https://github.com/MikeSchulze/gdUnit4 and that one looks much more promising. I will raise a PR soon to add these tests!

bitbrain:
Added a first set of unit tests here: https://github.com/bitbrain/beehave/pull/105

PR #6, opened 2022-07-29
disable physics process on root disable

JoelPagliuca:
## Description

disable/enable `physics_process` with `enable()` / `disable()` to save some cycles (especially with lots of AI agents running with disabled BTs)

I think we've both ended up inspired by the same BT implementations, I have some slight differences in my implementation so I thought I'd send some PRs to see what you thought, huge fan of the videos btw!

## Screenshots

image

## Checklist

- [x] Unit tests (Godot 4 only after https://github.com/bitbrain/beehave/issues/2 is done)


bitbrain (approved):

bitbrain:
Nice one!

PR #9, opened 2022-08-04
Allow custom actor on the BeehaveRoot node

Donorhan:
## Description

Depending on the project it happens that the parent node is not the actor node. 

For example in my case I have the following architecture:
 - PlayerController
    - Character

Or

- AIController (AI type)
  - Character
  - BehaviorRoot node

The purpose of this parameter is to be able to use a node other than the parent node, if the parameter is not specified the code fallback on the default behavior : using the parent node.

## Addressed issues

- Manage nodes in a more flexible way

## Screenshots

Capture d’écran, le 2022-08-04 à 18 50 13

## Checklist

### DIFF
--- a/addons/beehave/nodes/beehave_root.gd
+++ b/addons/beehave/nodes/beehave_root.gd
@@ -9,6 +9,10 @@ const RUNNING = 2
 
 @export var enabled : bool = true
 
+@export_node_path var actor_node_path : NodePath
+
+var actor : Node
+
 @onready var blackboard = Blackboard.new()
 
 func _ready():
@@ -16,26 +20,31 @@ func _ready():
 		push_error("Beehave error: Root should have one child")
 		disable()
 		return
+
+	actor = get_parent()
+	if actor_node_path:
+		actor = get_node(actor_node_path)
+
 	set_physics_process(enabled)
 
 func _physics_process(delta):
 	blackboard.set("delta", delta)
 
-	var status = self.get_child(0).tick(get_parent(), blackboard)
-	
+	var status = self.get_child(0).tick(actor, blackboard)
+
 	if status != RUNNING:
 		blackboard.set("running_action", null)
-	
+
 func get_running_action():
 	if blackboard.has("running_action"):
 		return blackboard.get("running_action")
 	return null
-		
+
 func get_last_condition():
 	if blackboard.has("last_condition"):
 		return blackboard.get("last_condition")
 	return null
-	
+
 func get_last_condition_status():
 	if blackboard.has("last_condition_status"):
 		var status = blackboard.get("last_condition_status")


bitbrain:
@Donorhan great change! Could you raise a similar PR for the 3.x branch, too?

bitbrain (approved):

Donorhan:
@bitbrain Yes of course 🙂

kaluluosi:
@Donorhan 
I'm very curious about your playercontroller aicontroller. Are you designed in imitation of UE4's gameplay framework? Is the playercontroller a node or a kinematical body? Can you share it?

Donorhan:
@kaluluosi I never used the unreal engine, I mostly worked on Unity, so may be 😅

I have a Character node which is a simple character (CharacterBody2D) with "action" functions like move to a point, take an object, attack, etc. This node also manages its animations, attributes (life points, inventory), .... The idea is to see the Character node as a simple puppet that you can control.

Next to that I have two other nodes (extending Node2D) that have a Character node as reference:
- PlayerController: this node listens for user input and gives commands to the puppet
- AIController: this node is composed of BehaviorTree and applies actions to the Character node linked to it

With this architecture the control logic is decoupled from the character logic. A character can be controlled by an AI or a user and this can change during the game.

I can make an example with code if you want more details 🙂

ai_node
character_node

kaluluosi:
> @kaluluosi I never used the unreal engine, I mostly worked on Unity, so may be 😅
> 
> I have a Character node which is a simple character (CharacterBody2D) with "action" functions like move to a point, take an object, attack, etc. This node also manages its animations, attributes (life points, inventory), .... The idea is to see the Character node as a simple puppet that you can control.
> 
> Next to that I have two other nodes (extending Node2D) that have a Character node as reference:
> 
> * PlayerController: this node listens for user input and gives commands to the puppet
> * AIController: this node is composed of BehaviorTree and applies actions to the Character node linked to it
> 
> With this architecture the control logic is decoupled from the character logic. A character can be controlled by an AI or a user and this can change during the game.
> 
> I can make an example with code if you want more details 🙂
> 
> ai_node
> 
> character_node



Awesome! @Donorhan 
Recently, I'm also designing a class that abstracts the player controller and treats the kinematicbody as a puppet.
I always thinking about how to implement control switch like Super Mario Odyssey.
image

Most tutorial teach us to deal player Input inside the character(KinematicBody), I think it's not a good way.
Official Docs say that we should deal Input in _input/_unhandle_input, but I never see the usecase.

So I imitate the idea of UE4's Gameplay Framework.
We need a class abstract player controller to handle input and possess characters.
AIController possess enemy. If we want , we can possess a enemy, then we take place the AIController with PlayerController.
Even make p1 p2 controller easy. Easy to implement character choose system.

But I didn't finish it yet.
My mind is blocked, and I need to absorb other people's ideas to help me get enlightened.
Would you like to show me how you implement it?

bitbrain:
@Donorhan @kaluluosi feel free to create a discussion here https://github.com/bitbrain/beehave/discussions instead as the pull requests are not meant for Q&A conversations unrelated to the pull request itself.

Issue #1, opened 2022-07-18
Debug view in editor


bitbrain:
**Is your feature request related to a problem? Please describe.**
When running games with behavior trees it is difficult to follow what is currently happening within it. Setting breakpoints within individual nodes is not ideal and when there is a bug it is quite difficult to figure out which nodes of the tree are currently running and what status they are currently in.

**Describe the solution you'd like**
Introduce a visual view inside the editor when clicking on a behavior tree that renders the behavior tree visually and highlights the status of every node in the tree and the "flow" animated as well.

## Example from **Tomate_Salat**

https://user-images.githubusercontent.com/822035/215510419-c6990290-cd60-440d-bb90-ae6d2ec52577.mp4

GammaGames:
Could the debug view be drawn with the `_draw` function by the root node? It'd be easy to enable/disable "debugging" with the node's `visible` property (somehow it'd have to default to disabled), and you could draw the behavior trees in the game world as the game runs. Would be cool to see the states highlight and change

bitbrain:
@GammaGames that is an interesting idea. I was more thinking about having it as a dedicated node scene (with Godot 4 you could have it as a separate window next to your game).

The problem with showing it "in the world" is that especially for pixel art games, the camera zoom makes it very difficult to properly show the behaviour tree.

GammaGames:
I agree that a separate scene would be easier, and you could always throw it inside a viewport to render it in game

And yay, separate windows!

Wichamir:
What's the state of this? Debugging without this is really painful.

bitbrain:
@GammaGames worked on a [Proof of Concept in the past](https://gist.github.com/GammaGames/28f3368128b67a4c02af0faed679cc93) (see https://github.com/GammaGames/beehave/pull/1). Unfortunately, that solution was flawed due to the following:

- someone using this addon should not have to setup additional stuff to get debugging to work. It should _just work_
- there were some API breaking changes we had to do in order to cater for the debugging functionality. With the `4.x` branch we may get away with it (since Godot 4 is officially not out yet)

The important thing is that there needs to be three things to make this work:

1. a dedicated debug view inside the Godot editor (perhaps in a new tab or when clicking on a behaviour tree node?) that displays the whole behaviour tree and shows the status of each node (e.g. by using colours, @GammaGames did a good job at that in his first attempt)
[...]

PR #11, opened 2022-08-06
Allow user to set process mode

GammaGames:
## Description

I added the option to set `process_node` to the `BeehaveRoot` script. It has a few options:
* `PHYSICS_PROCESS`: Current behavior
* `IDLE`: Process ticks in `_process`
* `MANUAL`: Call the tree's `tick` method manually

(This is mostly useful for the debug tools I'm working on, but it was easy to add so I did it in a separate PR)

## Addressed issues

- Prep for #1, will allow the user to manually call ticks to step through the code

## Screenshots

image

## Checklist

### DIFF
--- a/addons/beehave/nodes/beehave_root.gd
+++ b/addons/beehave/nodes/beehave_root.gd
@@ -10,7 +10,8 @@ const RUNNING = 2
 enum PROCESS_MODE {
 	PHYSICS_PROCESS,
 	IDLE,
-	INPUT
+	INPUT,
+	MANUAL
 }
 
 export (bool) var enabled = true
--- a/addons/beehave/nodes/beehave_root.gd
+++ b/addons/beehave/nodes/beehave_root.gd
@@ -13,7 +13,7 @@ enum PROCESS_MODE {
 	MANUAL
 }
 
-export (PROCESS_MODE) var process_mode = PROCESS_MODE.PHYSICS_PROCESS
+export (PROCESS_MODE) var process_mode = PROCESS_MODE.PHYSICS_PROCESS setget set_process_mode
 export (bool) var enabled = true
 
 onready var blackboard = Blackboard.new()
@@ -61,14 +61,17 @@ func get_last_condition_status():
 			return "RUNNING"
 	return ""
 
-
 func enable():
 	self.enabled = true
 	set_process(process_mode == PROCESS_MODE.IDLE)
 	set_physics_process(process_mode == PROCESS_MODE.PHYSICS_PROCESS)
 
-
 func disable():
 	self.enabled = false
 	set_process(self.enabled)
 	set_physics_process(self.enabled)
+
+func set_process_mode(value):
+	process_mode = value
+	set_process(process_mode == PROCESS_MODE.IDLE)
+	set_physics_process(process_mode == PROCESS_MODE.PHYSICS_PROCESS)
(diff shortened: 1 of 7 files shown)


GammaGames:
Followed by: https://github.com/GammaGames/beehave/pull/1

https://user-images.githubusercontent.com/7832163/183264739-7c033b6e-08a2-4af1-aed0-1b458a01ab0a.mp4

bitbrain (approved):

bitbrain:
@GammaGames looks mostly good to me apart from some smaller comments. There are some merge conflicts it seems, though.

addons/beehave/nodes/beehave_root.gd
```diff
@@ -7,6 +7,13 @@ const SUCCESS = 0
 const FAILURE = 1
 const RUNNING = 2
```
bitbrain, line 9: Could you name this ProcessMode instead?

addons/beehave/nodes/beehave_root.gd
```diff
@@ -47,12 +61,17 @@ func get_last_condition_status():
...
 
 func disable():
 	self.enabled = false
-	set_physics_process(false)
+	set_process(self.enabled)
+	set_physics_process(self.enabled)
+
+func set_process_mode(value):
+	process_mode = value
+	set_process(process_mode == PROCESS_MODE.IDLE)
```
bitbrain: A bit of a funny one: if you enable the idle thread from within the physics thread there might be a race condition where it starts processing the tree already while this very method is still running. Be mindful of that edge case.

bitbrain (changes requested):
Address comments first.

addons/beehave/nodes/beehave_root.gd
GammaGames: Done!

addons/beehave/nodes/beehave_root.gd
GammaGames: Good catch! Should I add a note, or make it wait until the next tick to change the process mode? Might be able to do the latter with the signals from the debug PR

bitbrain: For now, add a FIXME comment. Let's see if someone has issues in future and we might implement a fix for it.

bitbrain (dismissed):

GammaGames:
Conflict should be fixed now!

addons/beehave/nodes/beehave_root.gd
```diff
@@ -56,12 +70,17 @@ func get_last_condition_status():
 			return "RUNNING"
 	return ""
 
-
 func enable():
 	self.enabled = true
-	set_physics_process(true)
-
+	set_process(process_mode == ProcessMode.IDLE)
```
bitbrain: One last thing: could we do this instead
here?

`set_proccess_mode(self.process_mode)`

addons/beehave/nodes/beehave_root.gd
```diff
@@ -25,26 +32,33 @@ func _ready():
 	if actor_node_path:
 		actor = get_node(actor_node_path)
 
-	set_physics_process(enabled)
+	set_process(enabled and process_mode == ProcessMode.IDLE)
```
bitbrain: Same here

addons/beehave/nodes/beehave_root.gd
GammaGames: Good idea!

bitbrain (approved):

PR #13, opened 2022-08-08
Add star icon to sequencer_star sequencer

creadicted:
## Description

I like to see the difference between a sequencer and a sequencer star node. I added an icon variant to the node. If this is acceptable, I can add one for selector as well. 
 

## Screenshots

example

## Checklist


bitbrain:
@creadicted thanks a bunch! Could we do a similar change for selector_star?

creadicted:
> @creadicted thanks a bunch! Could we do a similar change for selector_star?
selector_star

bitbrain (approved):

PR #14, opened 2022-08-09
Add signals to nodes

GammaGames:
## Description

This was primarily for assisting with debugging, but has a few notes:

1. The user should implement their own `_tick` method if they want the signals to fire. I did it this way to be consistent with the other methods like `_draw` to draw on the canvas, where you call `update` to force a redraw.
2. The user should still call `tick` on each child node to process them
3. If the user overrides the `tick` method accidentally nothing will break. This ensures old code still works (I've been testing it on your demo project)

## Addressed issues

- Related to #1 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Conditions are **leaf nodes** of type `ConditionLeaf`. They should be kept simpl
 class_name IsVisibleCondition
 extends ConditionLeaf
 
-func tick(actor, blackboard):
+func _tick(actor, blackboard):
     if actor.visible:
         return SUCCESS
     return FAILURE
@@ -60,7 +60,7 @@ Actions are **leaf nodes** of type `ActionLeaf`. They can be long running potent
 class_name MakeVisibleAction
 extends ActionLeaf
 
-func tick(actor, blackboard):
+func _tick(actor, blackboard):
     if actor.visible:
         return FAILURE
 	actor.visible = true
--- a/addons/beehave/nodes/beehave_node.gd
+++ b/addons/beehave/nodes/beehave_node.gd
@@ -4,11 +4,9 @@ class_name BeehaveNode, "../icons/action.svg"
 
 enum { SUCCESS, FAILURE, RUNNING }
 
-signal tick_start(node)
 signal tick_end(node, status)
 
 func tick(actor, blackboard):
-	emit_signal("tick_start", self)
 	var result = self._tick(actor, blackboard)
 	emit_signal("tick_end", self, result)
 	return result
--- a/addons/beehave/nodes/beehave_node.gd
+++ b/addons/beehave/nodes/beehave_node.gd
@@ -4,5 +4,14 @@ class_name BeehaveNode, "../icons/action.svg"
 
 enum { SUCCESS, FAILURE, RUNNING }
 
+signal tick_start(node)
+signal tick_end(node, status)
+
 func tick(actor, blackboard):
+	emit_signal("tick_start", self)
+	var result = self._tick(actor, blackboard)
+	emit_signal("tick_end", self, result)
+	return result
+
+func _tick(actor, blackboard):
 	pass
--- a/addons/beehave/nodes/beehave_root.gd
+++ b/addons/beehave/nodes/beehave_root.gd
@@ -22,8 +22,8 @@ var actor : Node
 
 onready var blackboard = Blackboard.new()
 
-signal tick_start(node)
-signal tick_end(node, status)
+signal tick_start(blackboard)
+signal tick_end(blackboard)
 
 func _ready():
 	if self.get_child_count() != 1:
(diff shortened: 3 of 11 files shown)


bitbrain:
A few initial thoughts on this:

- I like the idea of exposing signals for users to connect to
- You introduced a breaking change (renaming the tick method) which is not desired. People using Beehave and "just upgrading" would suddenly have lots of errors or even worse - might not even notice that the method got renamed (as it still just runs) and none of their code would work

Perhaps I am understanding it wrong what the purpose of this PR is after all.

GammaGames:
> You introduced a breaking change (renaming the tick method) which is not desired

I understand why a change this large isn't desired. This method seemed more proper since the nodes are emitting the signals themselves and users should already be used to overriding `_process` and `_physics_process` for "private" methods.

> might not even notice that the method got renamed (as it still just runs)

That is the point. It's less a breaking change and more of a "the old behavior is still consistent, but if you want signals (i.e. for debugging) add an `_` to the start of your tick method" change.

> would suddenly have lots of errors [...] none of their code would work

What errors are you expecting? The change was made specifically to ease compatibility since overriding the regular `tick` method would only stop the signals from firing. The tree is still calling `tick` on child nodes, none of the function signatures changed.

If you'd prefer to keep only the `tick` method I have a few possible solutions:

* Make the user manually emit the signals in every composite or action node by adding a few lines of boilerplate to the start and end of every custom tick method
* Have the parent composite nodes force the child to emit the signals (like my previous implementation: https://github.com/GammaGames/beehave/blob/debug/addons/beehave/nodes/composites/sequence.gd)

Does Godot 4 have decorators? That's essentially all the `tick` -> `_tick` calls are doing, manually "decorating" the function so it emits the signals.

bitbrain:
@GammaGames ah, gotcha now. Thanks for elaborating.

Just a general thought: firing signals might have an undesired performance impact so I am wondering if there is a mechanism to only fire them if people actually connect to them. Perhaps Godot does this out of the box already?

GammaGames:
Good question! I looked at the source code: https://github.com/godotengine/godot/blob/1ea8b9d2b7da2285e890e087ff2b354e4d458b66/core/object.cpp#L1167

It loops through all connections in a map and emits the signal for each. So in the event that there are no signals, it calls that method and skips down to the end. Considering it's C++ I don't expect it to have much impact, though I am now curious if the profiler could shed any light.

Looks like objects also have a way to block signals, just discovered this: [` set_block_signals`](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-method-set-block-signals)

We could somehow set it to false by default, then make any tools that rely on the signals turn it back on. I don't think anything is using the `_ready` method aside from the base tree

bitbrain:
@GammaGames Interesting... could you use the version of current `godot-3.x` and run some sample scene with the profiler on (no signals connected) and compare it with your branch? If there is no performance difference I'll give this a tick.

GammaGames:
First time using the profiler, hopefully this is useful. I took your demo from the youtube video and updated the addon, disabled the forester and trader AI, and let the lumberjack run until he filled up the stash and went home.

Without signals:
* Frame time:
image
* Average time:
image

With signals (and updated `tick` -> `_tick`):
* Frame time:
image
* Average time:
image

With only `tick_end` signal for `BehaveNode`, `BehaveRoot` still has both start and end:
* Frame time:
image
* Average time:
image

There is overhead! I don't think a signal for each node's start tick is necessary tbh, most uses would probably only want the end result.

GammaGames:
Also if you've got a demo for 4.x I'd be happy to make a separate PR with the changes. I haven't tried GDScript 2.0 yet, so it might be fun

bitbrain:
4ms average sounds not like a lot but especially on lower-end PCs this could mean a drastic FPS drop. To me, this sounds more like a problem of Godot Engine itself, as emitting signals should not even have any impact on FPS if there are 0 connections made to them. It looks like it does, though. Within other Godot games, signals are mostly emitted on particular events, so perhaps Godot is not meant to emit them every single frame (therefore the added delay).

For https://github.com/GammaGames/beehave/pull/1 to work, I am wondering if we could find a solution without relying on signals after all.

GammaGames:
I agree that it could add up, thankfully emitting the signal only at tick end seems to not have much overhead. IMO that's the best solution (using signals, I can't think of any other way!) and I updated the branch with those changes. I also pass the blackboard on tree root tick start and end, in case the user wanted to make sure variables are set for the tree's blackboard or want to do something based on the resulting blackboard.

Here's my current debug helper scene: https://gist.github.com/GammaGames/28f3368128b67a4c02af0faed679cc93/fd527d0ada7e8a0749c2c6c5aab6d3c7d4a06820

Alternatively: we might be able to cache each node's status in the blackboard with `blackboard.set(self, result)` before returning in the `tick` method, but you'd need some way to remember the path of traversed nodes in a given tick. Appending to an array might work, but idk how much extra performance *that* would have.

GammaGames:
Decided to try that option out:
image

Might be because I'm running low on battery. WIll test again tomorrow with less stuff running and plugged in: https://github.com/GammaGames/beehave/tree/signals-alt

And the updated debug script, to see how the changes are used: https://gist.github.com/GammaGames/28f3368128b67a4c02af0faed679cc93/fb8e4fc7932dc599cb61661a8aee3424dfc19f78

GammaGames:
Using the `signals-alt` implementation:
* Frame time:
image
* Average time:
image

I think it's the cleanest implementation (only emitting two signals total), but the updating of the blackboard for each node and emitting every tick has some overhead.  
This is all on a Quad-Core Intel® Core™ i5-8400H CPU @ 2.50GHz btw.

bitbrain:
Closing this for now because I do not see much purpose for this apart for the debug UI itself. Perhaps in a future PR we can bring this back in combination with a debug UI that makes it in!

GammaGames:
I figured you would eventually, I haven't been able to think of a better way to debug. Don't know much about the editor features behind the scenes

Issue #26, opened 2022-08-23
failer.gd wrong indent


Foxomat:
The indent of the `return FAILURE` in `failer.gd `is wrong, it shouldn't be in the for-loop.

### Discussed in https://github.com/bitbrain/beehave/discussions/25

<div type='discussions-op-text'>

<sup>Originally posted by **Foxomat** August 22, 2022</sup>
In the failer.gd script, the last `return FAILURE` line is inside of the for loop, is that correct? If it has no children, it would return nothing.

failer.gd
```
extends Decorator

class_name AlwaysFailDecorator, "../../icons/fail.svg"


func tick(action, blackboard):
	for c in get_children():
		var response = c.tick(action, blackboard)
		if response == RUNNING:
			return RUNNING
		return FAILURE
```
</div>

PR #27, opened 2022-08-23
fixed wrong indent of `return FAILURE` statement in `failer.gd`

Foxomat:
## Description & Addressed issues

see #26 

## Checklist


bitbrain (approved):

Issue #30, opened 2022-08-31
The export var enabled is useless isn't it？


kaluluosi:
Godot version: x.x.x

**Describe the bug**
![image](https://user-images.githubusercontent.com/1620585/187610887-f6e8f328-8f21-4262-b68e-b285f50d0d16.png)

wether you set enabled true or false，the code doesn't use enabled to enable/disable behavior tree root. Tree still running.

**Expected behavior**
export var enable can disable behavior tree process on ready or dynamic disable process in runtime.

Nemesis-AS:
The `beehave_root.gd` does have `enable()` and `disable()` methods for enabling/disabling them during runtime. So this variable should not be modified at runtime.
The `_ready` function could be modified to enable/disable based on this variable

kaluluosi:
> The `beehave_root.gd` does have `enable()` and `disable()` methods for enabling/disabling them during runtime. So this variable should not be modified at runtime. The `_ready` function could be modified to enable/disable based on this variable

may be we can add setter function to set_enable.

bitbrain:
Closing this, as this issue is solved in Godot 4.

PR #35, opened 2022-09-07
Make export var enabled usefull

kaluluosi:
## Description
The logic of set_process_mode and enable/disable is confuse and conflict. 

```gdscript
func enable():
	self.enabled = true
	set_process_mode(self.process_mode)

func disable():
	self.enabled = false
	set_process(self.enabled)
	set_physics_process(self.enabled)

func set_process_mode(value):
	process_mode = value
	set_process(process_mode == ProcessMode.IDLE)
	set_physics_process(process_mode == ProcessMode.PHYSICS_PROCESS)
```

```gdscript
export (bool) var enabled = true  # it can't dynamic disable/enable in runtime, it just a hint var.
```
**In my mind, ProcessMode and enabled is just a flag.**
Process determine the logic by checking these flags.

## Addressed issues

- The export var enabled is useless isn't it？ #30

## Screenshots

```gdscript
func _process(delta):
	# check mode IDLE
	if process_mode == ProcessMode.IDLE:
		tick(delta)

func _physics_process(delta):
	# check mode PHYSICS_PROCESS
	if process_mode == ProcessMode.PHYSICS_PROCESS:
		tick(delta)

func tick(delta):
	# check enable
	if not enable:
		return

	blackboard.set("delta", delta)

	var status = self.get_child(0).tick(actor, blackboard)

	if status != RUNNING:
		blackboard.set("running_action", null)
```

Remove disable/enable,set_process_mode functions. No need any more.

Now it can dynamic set process_mode and enabled var to switch mode and disable it.
You can turn on/off enabled on inspector and it perform as set enabled by code.
And the value of var is sync to what it realy is.
The logic is clearly not confuse.

## Checklist


addons/beehave/nodes/beehave_root.gd
```diff
@@ -32,19 +32,23 @@ func _ready():
 	if actor_node_path:
 		actor = get_node(actor_node_path)
 
-	set_process_mode(self.process_mode)
 
 func _process(delta):
-	tick(delta)
-  
+	if process_mode == ProcessMode.IDLE:
```
bitbrain, line 37: Wouldn't Godot just not run the _process function anyways if the process_mode is not IDLE? (same for PHYSICS_PROCESS)

kaluluosi: > void set_process ( [bool](https://docs.godotengine.org/en/3.5/classes/class_bool.html#class-bool) enable )
> 
> Enables or disables processing. When a node is being processed, it will receive a [NOTIFICATION_PROCESS](https://docs.godotengine.org/en/3.5/classes/class_node.html#class-node-constant-notification-process) on every drawn frame (and the [_process](https://docs.godotengine.org/en/3.5/classes/class_node.html#class-node-method-process) callback will be called if exists). Enabled automatically if [_process](https://docs.godotengine.org/en/3.5/classes/class_node.html#class-node-method-process) is overridden. Any calls to this before [_ready](https://docs.godotengine.org/en/3.5/classes/class_node.html#class-node-method-ready) will be ignored.
> 

**How & Why**
At first. When we override _process/_physic_process , They enabled automatically.
So both _process and _physic_process is call by Godot every tick.
In _process function ,if process_mode not IDLE, _process still call, but it skip tick, because the condition check not pass.(Same to _physic_process)
Set process_mode to MANUAL or anyother int value(not IDLE or PHYSICS_PROCESS), it skip tick in both _process and _physic_process. Then you have to call tick by manual.

set_process/set_physic_process still can effect behaveroot's  _process/_physic_process. Same as the Pause Mode can effect it too.

**Performance** 
Well, switch mode or disable process by set_process may have good performance. Because it make Godot not to call beehaveroot's _process in the engine side. By flag, Godot still call _process every time.But behaveroot's _process or _physic_process is just simple condition check, not effect too much. And flag base is more easier implement.

addons/beehave/nodes/beehave_root.gd
```diff
@@ -69,16 +73,3 @@ func get_last_condition_status():
 			return "RUNNING"
 	return ""
 
-func enable():
```
bitbrain, line 72: We should not remove these methods. We still want to be able to manually enable or disable the behaviour tree programmatically.  Unless there is a different way to do this now?

bitbrain (changes requested):

bitbrain:
Closing this for now. Feel free to raise a new PR or re-open this one with the requested changes.

PR #36, opened 2022-09-08
Iconpaths

creadicted:
## Description
The way how icons are imported changed in godot 4
The support of the old way was now removed. 
## Addressed issues
https://github.com/godotengine/godot/pull/65264

## Checklist


bitbrain (approved):
LGTM

PR #48, opened 2022-09-17
Allow user to set process mode

creadicted:
## Description
I want to bring the change proposed and accepted here: 
https://github.com/bitbrain/beehave/pull/11/files#diff-fcaf318832e297e308d45827bfd3ccc205e55ec90d7af8a35d142cdc856e4918
to Godot 4

## Checklist
- [x ] changes performed compatible with Godot 4.x (raised in a separate PR against the `2.x` branch in case the change is not compatible with Godot 3.x like specific syntax)

### DIFF
--- a/addons/beehave/nodes/beehave_root.gd
+++ b/addons/beehave/nodes/beehave_root.gd
@@ -4,9 +4,21 @@ class_name BeehaveRoot
 @icon("../../icons/tree.svg")
 
 var Blackboard = load("res://addons/beehave/blackboard.gd")
+
 const SUCCESS = 0
 const FAILURE = 1
 const RUNNING = 2
+
+enum BeehaveProcessMode {
+	PHYSICS_PROCESS,
+	IDLE,
+	MANUAL
+}
+
+@export var beehave_process_mode: BeehaveProcessMode = BeehaveProcessMode.PHYSICS_PROCESS:
+	set(value):
+		beehave_process_mode = value
+		set_beehive_process_mode(value) 
 
 @export var enabled : bool = true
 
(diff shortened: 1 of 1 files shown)


addons/beehave/nodes/beehave_root.gd
```diff
@@ -4,10 +4,22 @@ class_name BeehaveRoot
 @icon("../../icons/tree.svg")
 
 var Blackboard = load("res://addons/beehave/blackboard.gd")
+
 const SUCCESS = 0
 const FAILURE = 1
 const RUNNING = 2
 
+enum BeehaveProcessMode {
```
bitbrain, line 12: Is it possible to call this just `ProcessMode`?

addons/beehave/nodes/beehave_root.gd
```diff
@@ -25,10 +37,16 @@ func _ready():
 	actor = get_parent()
 	if actor_node_path:
 		actor = get_node(actor_node_path)
+		
+	set_beehive_process_mode(self.beehave_process_mode)
```
bitbrain, line 41: probably just `set_process_mode`

addons/beehave/nodes/beehave_root.gd
creadicted: [ProcessMode ](https://docs.godotengine.org/en/latest/classes/class_node.html?highlight=ProcessMode#enumerations) are reserved enums of "Node". Therefore, I can not use either set_process_mode or ProcessMode

bitbrain: Could we then just re-use that one instead of introducing a new one?

addons/beehave/nodes/beehave_root.gd
bitbrain: @creadicted FYI

bitbrain:
Closing this for now. Feel free to raise a new PR or re-open this one with the requested changes.

Issue #50, opened 2022-10-30
Blackboard class is kind of broken in Godot 4


Wichamir:
Godot version: v4.0.beta3.official

**Describe the bug**

**Set and get don't work**
Blackboard overrides built-in set and get functions. This makes Blackboard's set and get functions unusable. I can't call them, because built-in functions are called instead. For example:
![image](https://user-images.githubusercontent.com/53086728/198893245-8b4ef150-e94d-4815-b0af-404bbc8c1b85.png)
![image](https://user-images.githubusercontent.com/53086728/198893250-feb15aca-41f1-4c19-aafd-ba6e544ada8a.png)
It expects 2 arguments, although it can take 3. If I pass 2 parameters, then I can run the game, but the built-in set function gets called instead of Blackboard's, so nothing happens, because there's no such property as "example".

**Memory leak**
Blackboard inherits [Object](https://docs.godotengine.org/en/latest/classes/class_object.html#class-object), but it isn't freed, when BeehaveRoot gets deleted, so it causes a memory leak.
![image](https://user-images.githubusercontent.com/53086728/198893478-b758bbd4-0833-4f0e-a4a5-09ff01fed81f.png)
I think Blackboard should inherit [RefCounted](https://docs.godotengine.org/en/latest/classes/class_refcounted.html) instead.

bitbrain:
The main issue is [that Godot 4 introduced a `set` method](https://docs.godotengine.org/en/latest/classes/class_object.html#id4
) which causes this problem. For Godot 4, we probably have to rename the method to something like `set_value` to solve the issue.

PR #53, opened 2022-11-12
Fix Blackboard class

Wichamir:
## Description

Blackboard class is broken in Godot 4. This pull request renames the method names to not override the built-in ones and changes the base class to RefCounted, which is a an equivalent to Godot 3's Reference, so Blackboard is safely freed by the engine.

## Addressed issues

- #50 

## Checklist


bitbrain (approved):

Issue #51, opened 2022-10-30
Add documentation comments


Wichamir:
**Is your feature request related to a problem? Please describe.**
It would be nice to have documentation for the most important classes like BeehaveRoot, ActionLeaf etc. Especially for people, who are yet learning behavior trees.

**Describe the solution you'd like**
Godot 4 adds [documentation comments](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_documentation_comments.html), so adding docs would be very easy.

For example:
![image](https://user-images.githubusercontent.com/53086728/198894995-3016bca6-9aae-4501-b09c-cb2136f3f08d.png)

Results in:
![image](https://user-images.githubusercontent.com/53086728/198895014-d8ee89b1-bf40-4b7c-ba79-48077c67aac5.png)
![image](https://user-images.githubusercontent.com/53086728/198895053-6df6bb95-7445-4e00-806c-0380593b4a64.png)

bitbrain:
I love this!

lostptr:
Can I work on this? Seems easy enough to just dump the descriptions in the readme into the classes.

bitbrain:
@lostptr no need to ask! Just pick up whatever you want to work on. Any contribution is highly appreciated.

PR #59, opened 2022-11-25
Added documentation comments

lostptr:
## Description

I've added documentation comments for all classes except for the plugin.gd script. Some of the documentation texts were changed based on [here](https://www.behaviortree.dev/docs/learn-the-basics/BT_basics). English is not my first language so it would be wise to check if there's any typos or something. 

## Addressed issues

#51 - Add documentation comments

## Screenshots

image

You can see them when you hover or mouse but, if you just click on the nodes the description in the bottom will not update (it will show the description for the previous selected node). I had to click on Recent and then select the node for the description to show. Seems to be a godot bug. I'm using godot 4 beta 6.

Also, I've used the `` sytax for the status codes but godot does not render them as markdown syntax. VS Code does, so I think it's neat to keep it this way:

image

## Checklist

- [ ] ~changes performed compatible with Godot 3.x~ (this will only be done for godot 4.x)
- [X] changes performed compatible with Godot 4.x (raised in a separate PR against the `2.x` branch in case the change is not compatible with Godot 3.x like specific syntax)

### DIFF
--- a/addons/beehave/blackboard.gd
+++ b/addons/beehave/blackboard.gd
@@ -1,3 +1,5 @@
+## The blackboard is an object that can be used to store and access data between
+## multiple nodes of the behaviour tree.
 class_name Blackboard extends RefCounted
 
 var blackboard: Dictionary = {}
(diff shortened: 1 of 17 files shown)


bitbrain (approved):
Good job!

Issue #45, opened 2022-09-13
Interruptions


and3rson:
**Is your feature request related to a problem? Please describe.**
There doesn't seem to be a way to notify an action that it needs to perform some cleanup due to interruption (say, stop moving if the player becomes visible.)

This is open to discussion since I'm not sure what would be the best way to achieve this.

**Describe the solution you'd like**
I'd suggest to add some sort of `Interruptor` node (or `interrupt(priority)` method to the tree itself), so that when an interruption happens (say, an `Interruptor` node returns a `SUCCESS`, etc), nodes with lower priority that are in `RUNNING` state would have `interrupt` method called.
I'm still relatively new to behavior trees, so please let me know if there are better ways to achieve this.

**Describe alternatives you've considered**
Let's say we have a "move" action that builds a path and starts the motion process (which is handled outside of the tree - e. g. `monster.walk_to_position(xy)` method which starts a `Tween`).
Now, in case the player becomes visible, I'd like to stop the movement, which means I'll need to explicitly call `monster.stop_movement()` before attacking. This "leaks" part of starting/stopping the walking process from "move" action to "attack" action.

creadicted:
> et's say we have a "move" action that builds a path and starts the motion process (which is handled outside of the tree - e. g. `monster.walk_to_position(xy)` method which starts a `Tween`).
> Now, in case the player becomes visible, I'd like to stop the movement, which means I'll need to explicitly call `monster.stop_movement()` before attacking. This "leaks" part of starting/stopping the walking process from "move" action to "attack" action.

Hey! I understand your suggestion and there could be an `exit()` method that is called when ever a node returns `FAILURE`:

```gdscript
		if response != FAILURE:
		   ...
		else:
			c.exit()
			last_execution_index += 1
```
But I am missing a good example where this is necessary. The example you provide is a wrong implementation of a behavior. 
You should move with each tick in the direction of the position and not define in a behavior calls that will create sideffects.

bitbrain:
@creadicted there is a deeper issue here and I think I know what @and3rson is getting at here:

A node can be in `RUNNING` state (within a Sequence* or Selector*) but at the root of the tree is a condition that suddenly prevents the running node to be called any longer. As a result, that node will hang in `RUNNING` state (but effectively it should be interrupted automatically)

plataris:
First thanks for this addon.
I am new to Godot but i will try to add my ideas on this issue.

[...]

PR #66, opened 2022-12-12
Added interrupt method for long running actions

lostptr:
## Description

This PR adds an `interrupt` method for the tree nodes when a long running action need to be prematurely interrupted as well as logic in `Selector` and `Sequence` nodes to call this method.

## Addressed issues

#45

## Screenshots

Well, no screenshots but here's a sample project I used to test things.
[beehave_sample.zip](https://github.com/bitbrain/beehave/files/10204162/beehave_sample.zip)

### DIFF
--- a/addons/beehave/nodes/beehave_node.gd
+++ b/addons/beehave/nodes/beehave_node.gd
@@ -4,6 +4,11 @@ class_name BeehaveNode extends BeehaveTree
 @icon("../icons/action.svg")
 
 ## Executes this node and returns a status code.
-## This method must be overwritten. 
+## This method must be overwritten.
 func tick(actor: Node, blackboard: Blackboard) -> int:
 	return SUCCESS
+
+
+## Called when this node needs to be interrupted before it can return FAILURE or SUCCESS.
+func interrupt(actor: Node, blackboard: Blackboard) -> void:
+	pass
--- a/addons/beehave/nodes/beehave_root.gd
+++ b/addons/beehave/nodes/beehave_root.gd
@@ -32,16 +32,22 @@ func _ready() -> void:
 
 func _physics_process(delta: float) -> void:
 	blackboard.set_value("delta", delta)
-
 	var status = self.get_child(0).tick(actor, blackboard)
 
-	if status != RUNNING:
-		blackboard.set_value("running_action", null)
+	# Updates the current running action.
+	var running_action = get_running_action() if status == RUNNING else null
+	blackboard.set_value("running_action", running_action)
 
 
 func get_running_action() -> ActionLeaf:
-	if blackboard.has_value("running_action"):
-		return blackboard.get_value("running_action")
+	var node = get_child(0)
+	while node != null:
+		if node is Composite:
+			node = (node as Composite).running_child
+		elif node is ActionLeaf:
+			return node
+	
+	push_error("Beehave error: Could not find running action in tree root '%s'." % name)
 	return null
 
 
--- a/addons/beehave/nodes/composites/composite.gd
+++ b/addons/beehave/nodes/composites/composite.gd
@@ -4,6 +4,16 @@ extends BeehaveNode
 class_name Composite
 @icon("../../icons/category_composite.svg")
 
+
+var running_child: BeehaveNode = null
+
+
 func _ready():
 	if self.get_child_count() < 1:
 		push_error("BehaviorTree Error: Composite %s should have at least one child (NodePath: %s)" % [self.name, self.get_path()])
+
+
+func interrupt(actor: Node, blackboard: Blackboard) -> void:
+	if running_child != null:
+		running_child.interrupt(actor, blackboard)
+		running_child = null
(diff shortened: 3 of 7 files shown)


bitbrain:
@lostptr is there a particular reason you only support the interrupt logic on `Sequence` and `Selector` but not `SequenceStar` and `SelectorStar`?

lostptr:
So, this will be a little hard to explain 😅
Lets look at 3 examples: what happends with a `SequenceStar` , a `Sequence` and a `SequenceStar` that it is a child of a `Sequence`.

## SequenceStar
cas1

Frame 1: The player is near so the enemy will wait a little and then attack.
Frame 2: The player is far away now but, since this is a `SequenceStar`, it will only run the Wait action and will not check again if the player is near or not, thus no interruptions will occur.
Frame 3: Wait returns `SUCCESS` and so Attack starts and returns `RUNNING`.
Frame 4: Attack returns `SUCCESS`.

## Sequence
case2
Frame 1: The player is near so the enemy will wait a little and then attack.
Frame 2: The player is far away now and the `Sequence` will return `FAILURE` but it needs to interrupt the Wait action. It will check if there's a running_action in the blackboard and call the interrupt method on it. After that, the `Sequence` node can immediately return `FAILURE`.

## Sequence with a condition and a SequenceStar as children
case3

Frame 1: The player is near so the enemy will wait a little and then attack.
Frame 2: The player is still near and the Wait action has returned `SUCCESS` (sorry, the gif is wrong at this part) so now the Attack action starts and returns `RUNNING`.
Frame 3: The player is far away now and the `Sequence` will return `FAILURE` but it needs to interrupt the Attack action. It will check if there's a running_action in the blackboard and call the interrupt method on it. Even though the Attack is not a direct child of the `Sequence`, the `Sequence` is the one responsible for interrupting any running actions since it is the one that "produced" the interruption (hope it makes sense lol). It can access the Attack node through the blackboard and call `interrupt()`. After that, the `Sequence` node can immediately return `FAILURE`.

**HOWEVER!**
What if the running action is not an indirect child of the Sequence/Selector that "produced" the interruption? I guess my implementation needs to address that as well.

***_Edit: I think we would be better off storing the running child in the composite nodes rather than in the blackboard. Then we could call the `interrupt()` recursively._

bitbrain:
> What if the running action is not an indirect child of the Sequence/Selector that "produced" the interruption? I guess my implementation needs to address that as well.

Yes.

> Edit: I think we would be better off storing the running child in the composite nodes rather than in the blackboard. Then we could call the interrupt() recursively.

@lostptr a bit of context: the reason I originally stored it in the `Blackboard` was that it was quick to implement. However, the problem with that is that a developer could simply clear the blackboard and as a result change/break the internal logic of the nodes. If you have a better idea to store that data outside of the blackboard go ahead.

addons/beehave/nodes/beehave_node.gd
```diff
@@ -4,6 +4,11 @@ class_name BeehaveNode extends BeehaveTree
...
 
 ## Executes this node and returns a status code.
-## This method must be overwritten. 
+## This method must be overwritten.
 func tick(actor: Node, blackboard: Blackboard) -> int:
 	return SUCCESS
+
+
+## Called when this node needs to be interrupted before it can return FAILURE or SUCCESS.
+func interrupt(actor: Node, blackboard: Blackboard) -> void:
```
bitbrain, line 13: another thought: any reason this cannot be a signal instead?

`
signal interrupted(actor: Node, blackboard: Blackboard)
`

lostptr: I would use a signal if it was trying to call a parent from a child node but in this case its the opposite. As a rule of thumb, I use `parent -> child = method` and `child -> parent = signal`.

addons/beehave/nodes/beehave_root.gd
```diff
@@ -32,18 +32,8 @@ func _ready() -> void:
...
 func _physics_process(delta: float) -> void:
 	blackboard.set_value("delta", delta)
-
 	var status = self.get_child(0).tick(actor, blackboard)
 
-	if status != RUNNING:
-		blackboard.set_value("running_action", null)
-
-
-func get_running_action() -> ActionLeaf:
```
bitbrain, line 42: This method cannot be removed, as it is used actively. For example, I depend on it in my game to display debug information about the tree. To avoid breaking changes, you will need to find a way to re-introduce this.

bitbrain, line 42: This method cannot be removed, as it is used actively. For example, I depend on it in my game to display debug information about the tree. To avoid breaking changes, you will need to find a way to re-introduce this.

bitbrain (changes requested):

addons/beehave/nodes/beehave_root.gd
lostptr: Oh ok, I could not find a reference thats why I removed. But sure, it gets hard to debug without this. I will add it back.

bitbrain (approved):

bitbrain:
This will not be cherry-picked for Godot 3 as it touches too many nodes that have been already modified as part of the Godot 4 migration and would introduce regression failures.

Issue #77, opened 2022-12-17
Show editor warnings on wrong configuration


bitbrain:
Many nodes require child nodes to be present in order to function correctly. Display [configuration warnings inside the editor](https://docs.godotengine.org/en/latest/tutorials/plugins/running_code_in_the_editor.html#reporting-node-configuration-warnings) in case a node does not have its children setup correctly.

## Nodes that require children

- `BeehaveTree`
- any decorator
- any composite: composite nodes ideally should have at least two children configured. Otherwise they are not useful

PR #80, opened 2022-12-18
Show editor warnings on wrong configuration

Wichamir:
## Description

This PR adds editor warnings, when Beehave nodes are wrongly configured. They only work with `@tool` annotation so I added it to every script (with conditions to not execute AI, when in editor).

## Addressed issues

- #77 

### Example:

image

### DIFF
--- a/addons/beehave/nodes/beehave_tree.gd
+++ b/addons/beehave/nodes/beehave_tree.gd
@@ -1,4 +1,5 @@
 ## Base class for all parts of the behaviour tree.
+@tool
 class_name BeehaveTree extends Node
 @icon("../icons/icon.svg")
 
@@ -8,3 +9,12 @@ enum {
 	FAILURE,
 	RUNNING
 }
+
+
+func _get_configuration_warnings() -> PackedStringArray:
+	var warnings: PackedStringArray = []
+	
+	if get_children().any(func(x): return not (x is BeehaveNode)):
+		warnings.append("All children of this node should inherit from BeehaveNode class.")
+	
+	return warnings
--- a/addons/beehave/nodes/composites/composite.gd
+++ b/addons/beehave/nodes/composites/composite.gd
@@ -1,7 +1,6 @@
 ## A Composite node controls the flow of execution of its children in a specific manner.
-extends BeehaveNode
-
-class_name Composite
+@tool
+class_name Composite extends BeehaveNode
 @icon("../../icons/category_composite.svg")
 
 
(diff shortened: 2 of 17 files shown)


addons/beehave/nodes/composites/composite.gd
```diff
@@ -1,18 +1,29 @@
 ## A Composite node controls the flow of execution of its children in a specific manner.
-extends BeehaveNode
-
-class_name Composite
+@tool
+class_name Composite extends BeehaveNode
```
bitbrain, line 3: One liners are now possible? WOW!

addons/beehave/nodes/beehave_tree.gd
```diff
@@ -8,3 +9,12 @@ enum {
 	FAILURE,
 	RUNNING
 }
+
+
+func _get_configuration_warnings() -> PackedStringArray:
+	var warnings: PackedStringArray = []
+	
+	if get_children().any(func(x): return not (x is BeehaveNode)):
```
bitbrain, line 17: This would not give a warning if `get_children()` is empty, right? We need to cover that case, too.

addons/beehave/nodes/composites/composite.gd
```diff
@@ -1,18 +1,29 @@
...
+		return
+	
 	if self.get_child_count() < 1:
 		push_error("BehaviorTree Error: Composite %s should have at least one child (NodePath: %s)" % [self.name, self.get_path()])
 
 
+func _get_configuration_warnings() -> PackedStringArray:
+	var warnings: PackedStringArray = super._get_configuration_warnings()
+	
+	if get_children().filter(func(x): return x is BeehaveNode).size() < 2:
```
bitbrain, line 21: Does this give a warning when `get_children()` is empty?

Wichamir: If there are no children `filter` returns empty array, so the condition is still correct.

addons/beehave/nodes/composites/composite.gd
Wichamir: Actually this was possible back in 3.x. It's just not that many people knew about it ;)

addons/beehave/nodes/beehave_tree.gd
Wichamir: Yeah, but I think it depends on Beehave node type. For example Leaf nodes **should** have no children, while composites should have at least 2 and decorators exactly 1 etc., so I implemented children count check separately depending on script. This line just checks if all children have proper type (that is BeehaveNode, otherwise it appends a warning).

bitbrain (approved):

bitbrain:

Issue #33, opened 2022-09-06
Bundle only addons/beehave folder in releases


bitbrain:
**Is your feature request related to a problem? Please describe.**
We currently bundle the entire .zip folder for the release but users of this addon do not care about all the extra files.

**Describe the solution you'd like**
Provide a better way of bundling releases to minimise the size.

kaluluosi:
well sir , I think the gitattributes file #31 already effect the  github release. You can try pre-publish a new one and test donwload source code zip, and you will found it only has addons file inside.

**But there is new problem, the LICENSE is no under addons/behave folder.Source zip or plugin in Godot Asset Library not contains LICENSES in package. Maybe we should make a copy under addons/behave folder.** 

_It's better make a copy LICENSE under addons/behave, it make sure it won't be override by other plugin's LICENSE._

bitbrain:
This worked flawlessly in #83! Closing this.

PR #83, opened 2022-12-19
Bundle only addons/beehave and script_templates in releases

Wichamir:
## Description

This works for godot-3.x branch already, but there's no .gitattributes file for godot-4.x, so I added it here too. Now it should also include the script_templates directory. I also copied the LICENSE to addons/beehave as advised in #33

## Addressed issues

- #33


bitbrain (approved):
Amazing!

Issue #74, opened 2022-12-16
Random composite task


d-bucur:
**Is your feature request related to a problem? Please describe.**
It would be nice to have a node that runs a random child to simulate some non deterministic behaviour (ie. an NPC in an RPG choosing randomly between walking around or talking to another NPC)

An example implementation can be found here:
https://github.com/libgdx/gdx-ai/wiki/Behavior-Trees#random-selector
https://github.com/libgdx/gdx-ai/wiki/Behavior-Trees#random-sequence

**Describe the solution you'd like**
Two new nodes added:
- Random selector: selects one of the children randomly, runs it and the composite returns the same state 
- Random sequence: shuffles children randomly, and executes them in order the same way that sequence would do

Design issues to resolve:
Star nodes variants: Is there a need to support the **star** variants? Should the default behaviour just be the star version as re-entry into a currently running state is expected? I also have some other thoughts on the whole star behaviour that I will post in a separate discussion.
Probabilities: To support non uniform probabilities for children their weights would need to be added somewhere in the new nodes and correspond to existing children. But this is a problem that can be solved in a new release after the uniform model is implemented.

**Describe alternatives you've considered**
You could have a separate script that randomly shuffles the children and you would get the same behaviour as what I am describing. The problem with this is that it's not really plug and play and that it's not so easy to determine when a composite execution has already ended and you need to do a reshuffle for the next execution. And you have to create a child node with a randomness script attached, and I'm not sure if that child would mess up the execution in some way.
An example:
- SelectorNode
|- ShuffleScript: doesn't extend beehave node, but changes order of the children
|- WalkAroundAction
|- TalkAction

Maybe having it as a decorator node for selector and sequence is a better approach? Would only have a single new node, but it would need to change the composite behaviour, and I'm not sure that's possible.

**Additional context**
I can work on implementing this during the holidays if there is an agreement on the design and an interest in merging this feature

bitbrain:
I really like your idea. It could act as a "proxy" node of its children:

It executes a random child (based on a probability) but keeps running that child until said child is completed (its status is the one of the child)
```
RandomComposite (RUNNING)
  |_ Child A (RUNNING) <-- runs over multiple ticks
  |_ Child B
```
When a child succeeds, this node succeeds:
```
RandomComposite (SUCCESS)
  |_ Child A (SUCCESS)
  |_ Child B
```
When a child fails, this node fails:
```
RandomComposite (FAILURE)
  |_ Child A
[...]

PR #90, opened 2022-12-27
New nodes: SequenceRandom and SelectorRandom

lostptr:
## Description

I've created two new nodes:

### sequence_random `SequenceRandom`
Shuffles the children and execute them until all of them return `SUCCESS` or one of them fail.

Has two flags:
- `resume_on_failure`: If set, it will continue where it left off. For example, in a node with actions `A`, `B` and `C`. After shuffling, the sequence will be C-A-B. Imagine action A fails and the sequence stops. Next time this node is executed, it will start with B and end the execution.
- `resume_on_interruption`: Its the same as above but only if it is interrupted.

### selector_random `SelectorRandom`
Shuffles the children and execute them until all of them return `FAILURE` or one of them succeed.

## Addressed issues

#74 

## Screenshots
My test case was an Enemy that would run, wait and use one of the 2 attack patterns it has:
scene

All of them will take a certain amount of frames to be executed.

debug

## Remarks
Both nodes use the global random number generator so to control the randomness it should be called `randomize()` somewhere in your project or the seed should be set using `seed("my seed".hash())`, for example.

Also, the icons look a bit streched...just a bit tho. I dont think it too impactful. 😜


bitbrain:
@lostptr Would you mind also updating the README with documentation for the new nodes? Just follow the same structure as for the existing ones.

addons/beehave/nodes/composites/sequence_random.gd
```diff
@@ -0,0 +1,61 @@
...
+## Whether the sequence should start where it left off after a previous failure.
+@export var resume_on_failure: bool = false
+## Whether the sequence should start where it left off after a previous interruption.
+@export var resume_on_interrupt: bool = false
+
+## A shuffled list of the children that will be executed in reverse order. 
+var _children_bag: Array[Node] = []
+var c: Node
+
+
```
bitbrain, line 18: @lostptr could we provide an optional @export variable for the seed on both nodes? By default, it could be 'null' / not set and uses the current behaviour.

We will need that for predictable unit testing once we got the CI PR merged.

lostptr: It uses the global random number generator so you can set the seed anywhere in the game (preferably in a singleton for these kind of things). Adding an export variable allows for the seed to be set multiple times and change mid game which is not ideal. The seed should be set only once, thats why I suggesting the singleton.

bitbrain: > The seed should be set only once, that's why I suggesting the singleton.

This is for the developer of the game to decide though, there are use-cases to set the seed individually. Not giving control over the randomness seems an unnecessary restriction.

Configuring the seed on a per-node basis is an additive change and can be done later. It will not collide with your proposed changes at all, so I am happy for it to be implemented later when there is truly a need for it!

lostptr: The worst thing is that we cant pass our own `RandomNumberGenerator` to the shuffle method, [it always uses the default global generator](https://github.com/godotengine/godot/blob/5674fa04aeb44a0d9cf48073fe2be25d1b804acf/core/variant/array.cpp#L618).

addons/beehave/nodes/composites/sequence_random.gd
```diff
@@ -0,0 +1,61 @@
...
+func _get_reversed_indexes() -> Array[int]:
+	var reversed = range(_children_bag.size())
+	reversed.reverse()
+	return reversed
+
+
+## Generates a new shuffled list of the children.
+func _reset() -> void:
+	print("=== done looping ===")
```
bitbrain: Remove print

bitbrain:
@lostptr just one small comment - LGTM otherwise. Good job!

addons/beehave/nodes/composites/sequence_random.gd
lostptr: Oof

lostptr: @bitbrain Done!

bitbrain (approved):

bitbrain:

Issue #89, opened 2022-12-21
Track performance metrics inside Godot


bitbrain:
Godot recently introduced performance monitoring: https://docs.godotengine.org/en/latest/tutorials/scripting/debug/custom_performance_monitors.html

The idea is that we can track various metrics in Godot and it will be displayed inside the performance tab. Possible metrics we can track:

| Metric Name | Description | Implemented |
|---|---|---|
| `beehave/tree_id/process_time` | The total process time per frame tick for the whole behaviour tree | https://github.com/bitbrain/beehave/pull/103 |
| `beehave/node_count` | The number of total beehave nodes in the scene | not yet |
| `beehave/active_count` | Number of active behaviour trees in the scene | not yet |

bitbrain:
Feel free to suggest any other metrics that we could start tracking.

goudcode:
Performance per tree is almost finished. I will open a PR for this in the next comming days.

PR #103, opened 2023-01-22
Implement performance metrics BehaveTree Node

goudcode:
## Description

I've implemented the performance per tree in the Godot engine profiler.
Time is being showed in nanosecconds and tree will be automatically added and removed to the profiler.

It might be better to disable all the metric code when running in release mode, example: use `if OS.is_debug_build()`
Any ideas are welcome.

## Addressed issues
#89 

## Screenshots
afbeelding

### DIFF
--- a/addons/beehave/nodes/beehave_tree.gd
+++ b/addons/beehave/nodes/beehave_tree.gd
@@ -22,8 +22,8 @@ enum {
 
 var actor : Node
 
-var _tree_metric_name : String
-var _tree_metric_value : float = 0.0
+var _process_time_metric_name : String
+var _process_time_metric_value : float = 0.0
 
 func _ready() -> void:
 	if Engine.is_editor_hint():
@@ -40,6 +40,13 @@ func _ready() -> void:
 	actor = get_parent()
 	if actor_node_path:
 		actor = get_node(actor_node_path)
+		
+	# Get the name of the parent node name for metric
+	var parent_name = actor.name
+	_process_time_metric_name = "beehave/%s-%s/process_time" % [parent_name, get_instance_id()]
+	
+	# Register custom metric to the engine
+	Performance.add_custom_monitor(_process_time_metric_name, _get_process_time_metric_value)
 
 	set_physics_process(enabled)
 
@@ -59,7 +66,7 @@ func _physics_process(delta: float) -> void:
 		blackboard.set_value("running_action", null, str(actor.get_instance_id()))
 	
 	# Check the cost for this frame and save it for metric report
-	_tree_metric_value = (Time.get_ticks_usec() - start_time) / 1000.0
+	_process_time_metric_value = (Time.get_ticks_usec() - start_time) / 1000.0
 	
 
 func _get_configuration_warnings() -> PackedStringArray:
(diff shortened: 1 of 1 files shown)


addons/beehave/nodes/beehave_tree.gd
```diff
@@ -22,6 +22,8 @@ enum {
 
 var actor : Node
 
+var _tree_metric_name : String
```
bitbrain: we may want to give this a better name: as it is tracking the `process_time` of this tree, perhaps something like this:
```suggestion
var _process_time_metric_name : String
```
(we are inside the tree already so using "tree" is redundant)

addons/beehave/nodes/beehave_tree.gd
```diff
@@ -93,3 +100,22 @@ func enable() -> void:
 
 func disable() -> void:
 	self.enabled = false
+
+
+func _enter_tree() -> void:
+	# Get the name of the parent node name for metric
+	var parent_name = get_parent().name
+	_tree_metric_name = "beehave/process_time/%s(%s)" % [parent_name, get_instance_id()]
```
bitbrain: From a structural perspective, we may want to have multiple metrics as part of the same tree. Therefore, I'd group them differently:
```suggestion
	_tree_metric_name = "beehave/%s-%s/process_time" % [parent_name, get_instance_id()]
```

addons/beehave/nodes/beehave_tree.gd
```diff
@@ -93,3 +100,22 @@ func enable() -> void:
 
 func disable() -> void:
 	self.enabled = false
+
+
+func _enter_tree() -> void:
+	# Get the name of the parent node name for metric
+	var parent_name = get_parent().name
```
bitbrain: Using `get_parent()` here is wrong, as there are different ways to associate a node to a tree (see `actor_node_path` export property)
```suggestion
	var parent_name = actor.name
```

bitbrain (changes requested):
Very good start! Apart from a few smaller comments this looks generally good.

Could we also add additional metrics like described on #89?

| Metric Name | Description |
|---|---|
| `beehave/node_count` | The number of total beehave nodes in the scene |
| `beehave/active_count` | Number of active behaviour trees in the scene |

Also reminder, that there is no need to check for debug mode, as Godot will only query the monitors inside the editor anyways and ensures not to call these in production mode.

addons/beehave/nodes/beehave_tree.gd
```diff
@@ -93,3 +100,22 @@ func enable() -> void:
...
+	Performance.add_custom_monitor(_tree_metric_name, _get_tree_metric)
+
+
+func _exit_tree() -> void:
+	# Remove tree metric from the engine
+	Performance.remove_custom_monitor(_tree_metric_name)
+
+
+# Called by the engine to profile this tree
+func _get_tree_metric() -> float:
```
bitbrain: Oh, this method also should be renamed to something like this:
```suggestion
func _get_process_time_metric_value() -> float:
```

goudcode:
The metric is now registered in the ready() function instead of the _enter_tree().
The reason for this is that the actor was not known in that function.

bitbrain (approved):

PR #119, opened 2023-02-11
Fix: Adapt usage of typed Arrays for RC1

Werninator:
## Description

Godot 4 RC1 is stricter with its usage of typed arrays. This PR will adapt the usage of typed arrays so it won't throw errors anymore.


bitbrain:
Thanks a ton!

bitbrain (approved):

PR #121, opened 2023-02-11
Fix: Check child count before applying interrupt

Werninator:
## Description

Currently, an out of bounds error will happen when:
- there's no child node inside the BeehaveTree
- BeehaveTree.enabled = false


bitbrain:
Could you also add a test here? https://github.com/bitbrain/beehave/blob/godot-4.x/test/beehave_tree_test.gd

bitbrain (approved):

PR #122, opened 2023-02-11
Move Blackboard initialization to export line

Werninator:
## Description

If BeehaveTree.enabled = false, the blackboard property will always be null if not otherwise assigned

This PR is there to allow editing a blackboard even if the tree is deactivated - reducing the need to add ifs outside the BeehaveTree itself. Also, correct me if I'm wrong - if the Tree starts out disabled, will it ever have a Blackboard (if not assigned)?

### DIFF
--- a/addons/beehave/nodes/beehave_tree.gd
+++ b/addons/beehave/nodes/beehave_tree.gd
@@ -27,7 +27,7 @@ signal tree_disabled
 		return enabled
 
 @export_node_path var actor_node_path : NodePath
-@export var blackboard:Blackboard
+@export var blackboard:Blackboard = Blackboard.new()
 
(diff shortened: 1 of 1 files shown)


bitbrain:
Could you also add a test here? https://github.com/bitbrain/beehave/blob/godot-4.x/test/beehave_tree_test.gd

Werninator:
I added a test to `beehave_tree_test.gd` and found out it that my solution would introduce orphans to the BeehaveTree. Also, Blackboard seems to be available after enabling - to answer my question above.

Closing this as making blackboard available even for disabled trees seems more a hustle to just leave it and wrap access to the blackboard with if-checks.

PR #126, opened 2023-02-12
Added `enter()` and `exit()` methods for beehave nodes

### LINKED ISSUE #118: Nodes with `enter()` and `exit()` methods
## Overview
We could have `enter()` and `exit()` methods for when its necessary to do some setup or cleanup before or after the first and last ticks.
The `enter()` and `exit()` methods would be called before the first `tick()` and after the last, respectively.
For more information, checkout #117 

## Possible implementation
We create an `execute()` method inside `BeehaveNode` to be called by _**parent nodes only**_. This method will call `enter()`, `tick()` and `exit()` when needed as well as return the result of `tick()`.
```gdscript
func execute(actor: Node, blackboard: Blackboard) -> int:
	if not _has_entered:
		enter(actor, blackboard)
		_has_entered = true
	
	var result = tick(actor, blackboard)

	if result != RUNNING:
		exit(actor, blackboard)
		_has_entered = false

	return result
``` 

This way, we only need to change composite nodes to call `execute()` instead of `tick()`. Composites and leaf nodes _**should never**_ override the `execute()` method, only `tick()`.

lostptr:
## Description

Many things had to be changed:
- Added `enter()` and `exit()` methods in `BeehaveNode`.
- All composites and decorators now have to call their children's enter and exit.
- Adjusted some composites to use those methods in their logic now:
  - SequenceStar
  - SelectorStar
  - SequenceRandom
  - SelectorRandom
- Since it's the parent's responsability to call those methods, all tests needed to be adjusted so that a `BeehaveTree` is added as the root.
- Some tests needed adjustments because now decorators **will** throw errors if they have no children.
- A new test was added to validade the enter and exit methods.
  - A new test action was added to make those tests possible (`MockAction`)

## Addressed issues

#117

### DIFF
--- a/addons/beehave/nodes/beehave_node.gd
+++ b/addons/beehave/nodes/beehave_node.gd
@@ -31,11 +31,11 @@ func interrupt(actor: Node, blackboard: Blackboard) -> void:
 
 
 ## Called before the first time it ticks by the parent.
-func enter(actor: Node, blackboard: Blackboard) -> void:
+func before_run(actor: Node, blackboard: Blackboard) -> void:
 	pass
 
 
-## Called after the last time it ticks and returns 
+## Called after the last time it ticks and returns
 ## [code]SUCCESS[/code] or [code]FAILURE[/code].
-func exit(actor: Node, blackboard: Blackboard) -> void:
+func after_run(actor: Node, blackboard: Blackboard) -> void:
 	pass
--- a/addons/beehave/nodes/beehave_node.gd
+++ b/addons/beehave/nodes/beehave_node.gd
@@ -18,6 +18,7 @@ func _get_configuration_warnings() -> PackedStringArray:
 	
 	return warnings
 
+
 ## Executes this node and returns a status code.
 ## This method must be overwritten.
 func tick(actor: Node, blackboard: Blackboard) -> int:
@@ -26,4 +27,15 @@ func tick(actor: Node, blackboard: Blackboard) -> int:
 
 ## Called when this node needs to be interrupted before it can return FAILURE or SUCCESS.
 func interrupt(actor: Node, blackboard: Blackboard) -> void:
+	pass
+
+
+## Called before the first time it ticks by the parent.
+func enter(actor: Node, blackboard: Blackboard) -> void:
+	pass
+
+
+## Called after the last time it ticks and returns 
+## [code]SUCCESS[/code] or [code]FAILURE[/code].
+func exit(actor: Node, blackboard: Blackboard) -> void:
 	pass
--- a/addons/beehave/nodes/beehave_tree.gd
+++ b/addons/beehave/nodes/beehave_tree.gd
@@ -27,11 +27,23 @@ signal tree_disabled
 		return enabled
 
 @export_node_path var actor_node_path : NodePath
-@export var blackboard:Blackboard
+@export var blackboard:Blackboard:
+	set(b):
+		blackboard = b
+		if blackboard and _internal_blackboard:
+			remove_child(_internal_blackboard)
+			_internal_blackboard.free()
+			_internal_blackboard = null
+		elif not blackboard and not _internal_blackboard:
+			_internal_blackboard = Blackboard.new()
+			add_child(_internal_blackboard, false, Node.INTERNAL_MODE_BACK)
+	get:
+		return blackboard if blackboard else _internal_blackboard
 
 var actor : Node
 var status : int = -1
 
+var _internal_blackboard: Blackboard
 var _process_time_metric_name : String
 var _process_time_metric_value : float = 0.0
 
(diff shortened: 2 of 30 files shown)


addons/beehave/nodes/beehave_node.gd
```diff
@@ -27,3 +28,14 @@ func tick(actor: Node, blackboard: Blackboard) -> int:
...
 	pass
+
+
+## Called before the first time it ticks by the parent.
+func enter(actor: Node, blackboard: Blackboard) -> void:
+	pass
+
+
+## Called after the last time it ticks and returns 
+## [code]SUCCESS[/code] or [code]FAILURE[/code].
```
bitbrain, line 39: Maybe worth mentioning that this happens after it previously returned `RUNNING`

Also, what would happen on nodes that never return `RUNNING` here - would they call `enter()` and `exit()` every tick?

addons/beehave/nodes/beehave_node.gd
```diff
@@ -27,3 +28,14 @@ func tick(actor: Node, blackboard: Blackboard) -> int:
 ## Called when this node needs to be interrupted before it can return FAILURE or SUCCESS.
 func interrupt(actor: Node, blackboard: Blackboard) -> void:
 	pass
+
+
+## Called before the first time it ticks by the parent.
+func enter(actor: Node, blackboard: Blackboard) -> void:
+	pass
```
bitbrain, line 35: I am wondering if there would ever be a performance impact of having these methods. Technically, `pass` will not execute anything so I'd expect this to be optimised at runtime to not take up any additional ms to compute.

bitbrain:
Great PR! We may want to be a bit more specific about the naming of the methods, as it is currently not quite clear when `enter` and `exit` are called exactly. 

> Called before the first time it ticks by the parent.

This makes it sound that the `enter` method is called only **once** in its lifetime, however, we should probably be more specific here in the description.

Could you also add something to the `README.md` to describe the additional behaviour? Perhaps under "Actions and Conditions"

addons/beehave/nodes/beehave_node.gd
lostptr: Would not this be an over optimization though?  The engine also have a lot of node methods that can but are not usually overriden. If the goal is to make it as fast as possible we should be writing in c++. If this slows down considerably we could take a look.

bitbrain: Great point, I have not considered that!

lostptr:
> Great PR! We may want to be a bit more specific about the naming of the methods, as it is currently not quite clear when `enter` and `exit` are called exactly.

How about `before_run` and `after_run`?

> Could you also add something to the `README.md` to describe the additional behaviour?

Oh yes! Forgot about that

bitbrain:
@lostptr good suggestion. Mhhh. I was first thinking about before_tick and after_tick but that would technically be wrong, right?

lostptr:
Yes, that would be wrong. I was thinking of giving an example in the documentation but maybe that's too much...?

> Say we have a character that will walk periodically. It must pick a destination, play the walk animation, move towards the destination and go back to an idle animation. 
> 
> This is how it would look like:
> image

Something along the lines...

bitbrain:
@lostptr I think that's great! Also, sorry for the merge conflict <3

lostptr:
@bitbrain ok yeah this messes everything lol

Now the beehave tree root is a weird composite that just runs everything without a care. 
I can't test now. Like, I dont even know how to make the tests work again because of this:
```gdscript
for child in get_children():
    if child is BeehaveNode:
        status = child.tick(actor, blackboard)
```
Why is this needed?

Alternatively, that "execute" implementation would require very minimal changes because all of the code would be handled inside the node itself, not their parent. I substancially reduces coupling and make it much easier to create new composites since there will be no need to think where to put the `before_run` and `after_run` calls. It just calls `_execute` and **that** method would know what to do.

With the current implementation, we introduce some constraints:
- Every behaviour tree must have a `BeehaveTree` as the root (many tests did not had that).
- The `BeehaveTree` must have only one child.
- Decorators **must** have a child (many tests where using `Succeeder` and `Failer` without children) else, an error occours.
- The logic for calling `before_run` and `after_run` must be implemented by every `Composite` and `Decorator` (and the `BeehaveTree`) meaning that, for every new composite there must have logic to know when to call the appropriate method.

That's why I had to make so many changes :/

Now consider what would need to be established with the "execute" implementation:
- `Composite`, `Decorator` and `BeehaveTree` would need to call `_execute()` instead of `tick()` on their children.
- Don't ever overload `_execute`.
 
Thats it. 

In fact, none even **needs to know that this method exists!** It is only relevant when you create Composites or Decorators and even then, all you need to know is that you have to call `child._execute()` instead of `child.tick()`. 
I'm insisting in this because now I'm seeing that is the simplest and most future proof solution. The chances of someone getting confused are slim if they read the docs, as is the case with any other library. ¯\\\_(ツ)\_/¯

bitbrain:
@lostptr that was just an attempt to safely run its only child as it can happen that someone adds no children to a tree but disabling it may not be a good idea as folks may want to add child nodes at runtime. What you are describing though should be still correct: BeehaveTree should only ever have one child at max. Perhaps replacing the for loop with if check + get_child(0) might be better after all.

addons/beehave/nodes/beehave_tree.gd
```diff
@@ -84,17 +84,25 @@ func _physics_process(delta: float) -> void:
 	
 	blackboard.set_value("delta", delta, str(actor.get_instance_id()))
 	
-	for child in get_children():
-		if child is BeehaveNode:
-			status = child.tick(actor, blackboard)
+  if self.get_child_count() == 1:
```
bitbrain:
```suggestion
	if self.get_child_count() == 1:
```

bitbrain:
@lostptr thank you!

bitbrain (approved):

PR #129, opened 2023-02-13
:bug: dynamically create or release inner Blackboard based on assigning external blackboards

### LINKED ISSUE #127: Blackboards count as orphans if created dynamically
Godot version: 4.0-rc1

As hinted from #122: It seems like, if a blackboard will be initialized at runtime (when there is no Blackboard assigned), it will count as an orphaned node, as it isn't added to the SceneTree.

I added following to `beehave_tree.gd:_ready()`:

```swift
if not blackboard:
    blackboard = Blackboard.new()
    blackboard.name = "Blackboard that has been created at runtime"

print_orphan_nodes()
```
Output:
```
Output: 52714014507 - Stray Node: Blackboard that has been created at runtime (Type: Node)
```

GDUnit4 didn't show me this being an orphan and in general this might not be an issue per se, but it did make me search for why there suddenly was an orphaned node appearing per Instance using a BeehaveTree.

Current workaround is to add a Blackboard and assign it, otherwise you'll have to ignore it. A solution could be to add new Blackboards to the tree.

**To Reproduce**
1. Add BeehaveTree to the Scene without an assigned Blackboard in it

**Expected behavior**
No orphaned node

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: win11

bitbrain:
## Description

There is currently an issue where someone can just assign `blackboard` to another property and break this addon. This PR comes with several changes:

- the internal blackboard is not considered an orphan any longer but is properly attached to the scene tree
- the behaviour tree now hosts multiple children (this may break things but we can discuss this on here)
- when assigning/unassigning blackboards at runtime, beehave will take care of automatically creating a new blackboard. Currently, values stored in the "old" blackboard will not be "moved over" (might be okay?)

## Addressed issues

- closes #127

### DIFF
--- a/addons/beehave/nodes/beehave_tree.gd
+++ b/addons/beehave/nodes/beehave_tree.gd
@@ -30,28 +30,35 @@ signal tree_disabled
 @export var blackboard:Blackboard:
 	set(b):
 		blackboard = b
-		if blackboard and internal_blackboard:
-			remove_child(internal_blackboard)
-			internal_blackboard.free()
-			internal_blackboard = null
-		elif not blackboard and not internal_blackboard:
-			internal_blackboard = Blackboard.new()
-			add_child(internal_blackboard)
+		if blackboard and _internal_blackboard:
+			remove_child(_internal_blackboard)
+			_internal_blackboard.free()
+			_internal_blackboard = null
+		elif not blackboard and not _internal_blackboard:
+			_internal_blackboard = Blackboard.new()
+			add_child(_internal_blackboard, false, Node.INTERNAL_MODE_BACK)
+	get:
+		return blackboard if blackboard else _internal_blackboard
 
 var actor : Node
 var status : int = -1
-var internal_blackboard: Blackboard
 
+var _internal_blackboard: Blackboard
 var _process_time_metric_name : String
 var _process_time_metric_value : float = 0.0
 
 func _ready() -> void:
 	if Engine.is_editor_hint():
 		return
+	
+	if self.get_child_count() > 0 and not self.get_child(0) is BeehaveNode:
+		push_warning("Beehave error: Root %s should have only one child of type BeehaveNode (NodePath: %s)" % [self.name, self.get_path()])
+		disable()
+		return
 		
 	if not blackboard:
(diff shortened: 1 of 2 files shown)


addons/beehave/nodes/beehave_tree.gd
```diff
@@ -27,25 +27,31 @@ signal tree_disabled
...
-	
-	if self.get_child_count() != 1:
-		push_warning("Beehave error: Root %s should have one child (NodePath: %s)" % [self.name, self.get_path()])
-		disable()
-		return
 		
 	if not blackboard:
-		blackboard = Blackboard.new()
+		internal_blackboard = Blackboard.new()
+		add_child(internal_blackboard)
```
Werninator: ~~`add_child` is in context of the BeehaveTree, right? Will this cause any errors or warnings, since it only wants to have one child?~~ (nvm, it's adressed in the PR Description)

Could we add the child to the root, like the Autoloads?

addons/beehave/nodes/beehave_tree.gd
```diff
@@ -144,3 +146,7 @@ func _exit_tree() -> void:
 # Called by the engine to profile this tree
 func _get_process_time_metric_value() -> float:
 	return _process_time_metric_value
+	
+
+func _get_current_blackboard() -> Blackboard:
```
Werninator: Could be probably replaced by adding a get(): to blackboard, right?

bitbrain: Great suggestion

addons/beehave/nodes/beehave_tree.gd
bitbrain: Actually, add_child has an argument that adds it to the "inner scope" that I will utilise! 

https://docs.godotengine.org/en/latest/classes/class_node.html#class-node-method-add-child

Werninator: > Actually, add_child has an argument that adds it to the "inner scope" that I will utilise!
> 
> https://docs.godotengine.org/en/latest/classes/class_node.html#class-node-method-add-child

awesome!

PR #134, opened 2023-02-14
Add nodes for Blackboard operations

### LINKED ISSUE #132: Add nodes for Blackboard operations
### Discussed in #125

Add 4 predefined nodes for `Blackboard` operations:

- `BlackboardSetAction` - sets a key to a specified value
- `BlackboardEraseAction` - erases the key from the blackboard
- `BlackboardHasCondition` - returns `SUCCESS` if the key exists in the blackboard, otherwise `FAILURE`
- `BlackboardCompareCondition` - compares two values from the blackboard by using the specified comparison operator

We can use the built-in `Expression` class to make it possible to define values, which are not a `String`.

Wichamir:
## Description

Here is a general draft. Tests are on the way. We also have to look for some icons for these new nodes.

`BlackboardSetAction` and `BlackboardCompareCondition` both use the `Expression` class to which the blackboard is passed as a base instance. This way you can do some cool stuff like:

- define any core type as a value
image

- copy values from one key to another image

- call methods 
image

Of course this is only useful for small snippets of code, so that you don't have to define a separate script whenever you want to compare a value for example. That's why I didn't use `@export_multiline`. This is only meant to be used with **small expressions**.

## Addressed issues

- #132


GeroVeni:
I think we could use expressions for the blackboard keys too, since they might not always be `String` either.

addons/beehave/nodes/leaves/blackboard_compare.gd
```diff
@@ -0,0 +1,39 @@
...
+@export var right_operand: String = ""
+
+
+@onready var _left_expression: Expression = parse_expression(left_operand)
+@onready var _right_expression: Expression = parse_expression(right_operand)
+
+
+func tick(actor: Node, blackboard: Blackboard) -> int:
+	var left: Variant = _left_expression.execute([], blackboard)
+	var right: Variant = _right_expression.execute([], blackboard)
```
GeroVeni: I think we could check for `has_execute_failed` here like in the `blackboard_set.gd`.

addons/beehave/nodes/leaves/blackboard_compare.gd
```diff
@@ -0,0 +1,39 @@
...
+	LESS_EQUAL,
+}
+
+
+@export var left_operand: String = ""
+@export_enum("==", "!=", ">", "<", ">=", "<=") var operator: int = 0
+@export var right_operand: String = ""
+
+
+@onready var _left_expression: Expression = parse_expression(left_operand)
```
bitbrain: Interesting, that `@export var expression: Expression` does not work. I'd love some sort of custom UI in Godot that would allow us to do that.

addons/beehave/nodes/leaves/blackboard_set.gd
```diff
@@ -0,0 +1,18 @@
...
+@export var value: String = ""
+
+
+@onready var _expression: Expression = parse_expression(value)
+
+
+func tick(actor: Node, blackboard: Blackboard) -> int:
+	var result: Variant = _expression.execute([], blackboard)
+	if _expression.has_execute_failed():
+		return FAILURE
```
Wichamir, line 21: Returning FAILURE here can be really confusing. We should always expect that the expression executes successfully. I think I will add assert or push_error here instead, to clearly inform the user that something went wrong.

Wichamir:
> I think we could use expressions for the blackboard keys too, since they might not always be `String` either.

We have to somehow make it clear that this is an expression and not a normal string though. I think many users will just casually input `some_key` instead of `"some_key"` and get confused why it doesn't work. Actually the same thing applies for other things in this PR. Some exported variables are simple Strings and others are Strings parsed into Expressions. Maybe we should name expression strings with `_expression` at the end?

addons/beehave/nodes/leaves/blackboard_set.gd
bitbrain: @Wichamir from a logical perspective, returning `SUCCESS` here would be wrong. Keeping it as it is makes sense to me. However, rather than logging or push error we should ask ourselves what could make the execution fail. I'd assume a wrong expression or invalid syntax of some sort? (reading the documentation suggests that https://docs.godotengine.org/en/latest/classes/class_expression.html#class-expression-method-execute)

In that case, we could even set up a configuration warning on these nodes in case someone setup a wrong expression syntax? What other reasons could there be for execute to fail?

bitbrain:
> We have to somehow make it clear that this is an expression and not a normal string though

is there no way for us to have some sort of "expression editor"? We could for example come up with our own (in a separate issue ticket, not here) and then write up a Godot proposal even that would allow people to do this:

```
@export var expression: Expression
```
which then will not show an editable text field but actually some sort of custom expression editor (with syntax highlighting) that would prevent saving invalid expressions.

addons/beehave/nodes/leaves/blackboard_set.gd
Wichamir: A configuration warning is a great idea! I can introduce it. This would eliminate syntax errors, but the problem still remains though. For instance:
`get_value("destination").rotated(PI)`
This is a perfectly valid syntax, but what if somehow the destination is not a Vector2, but a Vector3. Vector3 requires two arguments for the rotated method, so we get a runtime error.

Wichamir:
> is there no way for us to have some sort of "expression editor"?

There's [this plugin](https://github.com/don-tnowe/godot-inspector-extender), which introduces some nice inspector features using only GDScript, but it seems kind of hacky. We should probably look into [EditorProperty](https://docs.godotengine.org/en/stable/classes/class_editorproperty.html).

Maybe for now we can use `@export_placeholder`?

GeroVeni:
This is how it is currently done in the `AnimationNodeStateMachineTransition` in Godot. They have a `StringName` for the `advance_condition` and a multiline `String` for the `advance_expression`.
image

Wichamir:
@bitbrain @GeroVeni 

Pushed some changes:
- now everything is an expression
- `@export_placeholder` is used wherever the exported `String` is meant to be an expression
- configuration warnings show up if you input an invalid expression
- added documentation comments

I'm also thinking about moving blackboard-related code from Leaf to something like `BlackboardLeaf` (similarly to `BlackboardNode` in #135), but then I think I should rename everything for consistency (`BlackboardSetAction` => `BlackboardSetLeaf` etc.) (or not?).

bitbrain:
@Wichamir these are technically actions, so I like the idea of calling them *Action and *Condition. Leaf is a more generic tree concept that within beehave is still useful but for a user of a behaviour tree might be more confusing (as they will mostly always think about actions and conditions)

addons/beehave/nodes/leaves/blackboard_set.gd
Wichamir: @bitbrain Okay so in the end I left only `return FAILURE` here, because it turns out `Expression` class itself pushes an error message anyway if something goes wrong.

Wichamir:
> these are technically actions, so I like the idea of calling them *Action and *Condition. Leaf is a more generic tree concept that within beehave is still useful but for a user of a behaviour tree might be more confusing (as they will mostly always think about actions and conditions)

@bitbrain Yeah, you're right. I left it as it is.

Anyway, I added unit tests and I think it's pretty much ready for review. My only concern is if it won't be a nightmare to debug, but I guess we'll have to see for ourselves.

test/nodes/leaves/conditions/blackboard_has_test.gd
```diff
@@ -0,0 +1,45 @@
...
+
+func before_test() -> void:
+	blackboard_has = auto_free(load(__source).new())
+	actor = null
+	blackboard = auto_free(load(__blackboard).new())
+	blackboard_has.key = "\"%s\"" % [KEY]
+	runner = scene_runner(blackboard_has)
+
+
+func test_key_exists() -> void:
```
GeroVeni, line 28: Could we add one more test, with a valid but non-`String` key, e.g. an `int` or a `Vector2`

addons/beehave/nodes/leaves/blackboard_compare.gd
```diff
@@ -0,0 +1,55 @@
...
+	LESS_EQUAL,
+}
+
+
+## Expression represetning left operand.
+@export_placeholder(EXPRESSION_PLACEHOLDER) var left_operand: String = ""
+## Comparison operator.
+@export_enum("==", "!=", ">", "<", ">=", "<=") var operator: int = 0
+## Expression represetning right operand.
+@export_placeholder(EXPRESSION_PLACEHOLDER) var right_operand: String = ""
```
GeroVeni: I think it is important in the `BlackboardCompare` node (in cotrast to the other three) to have a bit more description on the operands. In the other nodes we have `key` and `value` fields which makes it clear what they are used for. In this node, however, they can use *any* two expressions. It would be nice if we expanded the docstring to include how they could use the value of existing keys e.g. 
*This value can be any valid GDScript expression. In order to use the existing blackboard keys for comparisson, use `get_value("key_name")` e.g. `get_value("direction").length()`*

GeroVeni (commented):
Really great work! I left a couple more comments but other than that look good to me! :)

test/nodes/leaves/conditions/blackboard_has_test.gd
Wichamir: Done.

bitbrain (approved):
Great work.

Werninator:
> I think many users will just casually input `some_key` instead of `"some_key"` and get confused why it doesn't work.

This just happened to me yesterday. Glad to know how they work now though, I'm gonna use them a lot :)

PR #135, opened 2023-02-14
add blackboard nodes

### LINKED ISSUE #132: Add nodes for Blackboard operations

GeroVeni:
**Closed** in favor of #134

Attempts to close #132

## Description

Adds the nodes described in #132.

Add 4 predefined nodes for `Blackboard` operations:

* `BlackboardSetAction` - sets a key to a specified value
* `BlackboardEraseAction` - erases the key from the blackboard
* `BlackboardHasCondition` - returns `SUCCESS` if the key exists in the blackboard, otherwise `FAILURE`
* `BlackboardCompareCondition` - compares two values from the blackboard by using the specified comparison operator

**Note:** This PR is still in progress, I just opened it in order to get some early feedback on it. It still has features that are missing such as:
1. Not tested yet
2. No unit tests added
3. No new icons for the blackboard nodes

### Initial implementation

I added a base `BlackboardNode` class that inherits from `Leaf` and added some functionality for parsing expressions (`parse_expression` in `blackboard_node.gd`). This is because I ended up reusing this logic in all of the other classes, in order to allow for non `String` keys. Currently it returns an array `[bool, Variant]` where the first value shows if there was an error during parsing the expression. I am not sure how much I like it this way, since it does not appear very GDScript like.

Another idea would be to have the `parse_expression` return only the parsed value (or null if it fails) and then have a second method in `BlackboardNode`, `has_parse_failed` that returns if the *last* parse operation failed. This is similar to how the built-in `Expression` already works.

Also, regarding the `BlackboardCompareCondition` node, the inequality conditions (`<, <=, >, >=`) are not guaranteed to be defined for all types and there is no check that the compare values can be compared. I do not know how we could check for that though.

### Thoughts on icons

For the icons I was thinking of using the icon used for the blackboard and change its color to the leaf color. This would be the icon for the `BlackboardNode` class. Then the subclasses will have that icon plus a "subscript" little icon at the lower right corner, an add symbol for the `BlackboardSetAction`, a minus symbol for the `BlackboardEraseAction`, a question mark for the `BlackboardHasCondition` and a less symbol (<) for the `BlackboardCompareCondition`.

### DIFF
--- a/addons/beehave/nodes/blackboard/blackboard_compare_condition.gd
+++ b/addons/beehave/nodes/blackboard/blackboard_compare_condition.gd
@@ -1,0 +1,61 @@
+## This node will compare two different keys from the blackboard, using the
+## specified [member comparison_type].
+@tool
+class_name BlackboardCompareCondition extends BlackboardNode
+
+## Defines the type of comparison to perform.
+@export var comparison_type: ComparisonType = ComparisonType.Less
+## The first blackboard key to compare.
+@export var blackboard_key_expression1: String
+## The second blackboard key to compare.
+@export var blackboard_key_expression2: String
+## If no [member blackboard_name] is specified, [code]"default"[/code] is going
+## to be used.
+@export var blackboard_name: String = ""
+
+enum ComparisonType {
+  Less,
+  LessOrEqual,
+  Greater,
+  GreaterOrEqual,
+  Equal,
+  NotEqual
+}
+
+func tick(actor: Node, blackboard: Blackboard) -> int:
+  var blackboard_name_ = "default"
+  if not blackboard_name.is_empty():
+    blackboard_name_ = blackboard_name
+
+  # Parse expressions and get the key values, checking for parse errors
+  var result1 = parse_expression(blackboard_key_expression1)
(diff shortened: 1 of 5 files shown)


GeroVeni:
Closed in favor of #134

Wichamir:
Oh sorry. Didn't know someone was already working on this. I should have commented about it myself. But I think we can cooperate on this. We still have to find some icons and write unit tests. And my code isn't nicely commented like yours is.

GeroVeni:
@Wichamir Feel free to take anything you like from my PR. I will look into prototyping the ideas for the icons and put them in the issue post for discussion

PR #138, opened 2023-02-15
:recycle: restructure sequence nodes + SequenceReactive node

### LINKED ISSUE #46: Restructure composite nodes to be less confusing
I have a sequence. Something like: 

- Get the closest object
- Walk to the closest object
- Interact with object

During the walk, the closest object can change. So how do I let the previous sequence nodes run if the 'Walk to the closest object' is `RUNNING `. From other behavior tree implementations, I learned about the _ReactiveSequence_. 

Type of ControlNode | Child returns FAILURE | Child returns RUNNING
-- | -- | --
Sequence | Restart | Tick again
ReactiveSequence | Restart | Restart
SequenceStar | Tick again | Tick again

* "Restart" means that the entire sequence is restarted from the first child of the list.
* "Tick again" means that the next time the sequence is ticked, the same child is ticked again. Previous sibling, which returned SUCCESS already, are not ticked again.
(Source: https://www.behaviortree.dev/sequencenode/)

A new node that follows the specification of the reactive sequence. 
If a leaf returns "Running" the previous children are ticked as well.

bitbrain:
## Description

Attempt to implement the following table:

Type of ControlNode | Child returns FAILURE | Child returns RUNNING
-- | -- | --
Sequence | Restart | Tick again
ReactiveSequence | Restart | Restart
SequenceStar | Tick again | Tick again

## Addressed issues

- closes #46


bitbrain:
@creadicted @Wichamir please let me know what you think about this! Note: I did not touch `Selector` yet but we may want to do something similar for that? What would the table look like for it?

addons/beehave/nodes/composites/sequence.gd
```diff
@@ -21,16 +27,28 @@ func tick(actor: Node, blackboard: Blackboard) -> int:
 		
 		match response:
 			SUCCESS:
+				successful_index += 1
 				c.after_run(actor, blackboard)
 			FAILURE:
-				# Interrupt any child that was RUNNING before.
-				interrupt(actor, blackboard)
```
bitbrain: Not 100% sure we need that? I guess we do?

bitbrain:
I will rework the tests so they reflect that table above.

Wichamir:
Just made a quick scene and tested these nodes. Sequence and ReactiveSequence both work according to the table. The SequenceStar however restarts in case of FAILURE, while it should tick again.

bitbrain:
@Wichamir I see what the problem is: https://github.com/bitbrain/beehave/blob/explicit-reactive-mode/addons/beehave/nodes/composites/sequence_star.gd#L33

I am not resetting the index when I return `FAILURE` but this happens in the `after_run` method: https://github.com/bitbrain/beehave/blob/explicit-reactive-mode/addons/beehave/nodes/composites/sequence_star.gd#L45

I should get rid of resetting anything there and only reset internally.

Wichamir:
> Note: I did not touch `Selector` yet but we may want to do something similar for that? What would the table look like for it?

I think originally this table was taken from BehaviorTree.CPP and the Selector node there is called [Fallback](https://behaviortree.github.io/BehaviorTree.CPP/FallbackNode/) and has two kinds:

Type of ControlNode | Child returns RUNNING
-- | --
Fallback | Tick again
ReactiveFallback | Restart

So in our case the current Selector would become SelectorReactive and the current SelectorStar would become Selector?

PR #157, opened 2023-02-25
Horizontal layout for Debug view

DarkMessiah:
## Description

Implement horizontal layout for Debug view

## Screenshots

https://user-images.githubusercontent.com/1756388/221362909-afe0f069-bc67-4aa0-a35a-e135cdb4b6e8.mp4


bitbrain:
Thank you!

addons/beehave/debug/tree_node.gd
```diff
@@ -190,23 +197,59 @@ func _center_nodes_between(left_node: TreeNode, right_node: TreeNode) -> void:
...
+func _calculate_y(node: TreeNode, offset: int) -> void:
+	node.y = offset
+	var sibling := node.get_most_left_sibling()
+	var max_size: int = node.item.size.y
+	while sibling != null:
+		max_size = max(sibling.item.size.y, max_size)
+		sibling = sibling.get_next_sibling()
+
+	for child in node.children:
+		_calculate_y(child, max_size + offset + LEVEL_DISTANCE * BeehaveUtils.get_editor_scale())
```
GeroVeni, line 255: I think this could be refactored to one function `_calculate_node_depth`. We could use another property like `layout_size` but reverse (maybe we can call it `node_height` ?) e.g.
```
var node_height: float:
	get:
		return size.x if horizontal else size.y
```
and use it in `var max_size: int = node.item.height` and `max_size = max(sibling.item.height, max_size)`.
Now the only difference between is the first line but we can solve that by assuming a vertical layout (so we use `node.y = offset`) and call `_calculate_node_depth` before we swap x and y, so in `update_positions` something like:
```
	_calculate_node_depth(self, 0)
	if horizontally:
		_swap_x_y(self)
```

Issue #156, opened 2023-02-22
Update to Godot 4-rc5


bitbrain:
Requires https://github.com/MikeSchulze/gdUnit4/issues/128 to be done first.

- also make sure to bump to rc5 in the Github action workflow.

PR #158, opened 2023-02-26
Update gdUnit4 to [166245a] for RC5 [6296b4600] compatibility

Werninator:
## Description

Obligatory gdUnit4 updat

## Addressed issues

- #156


bitbrain:
@Werninator you also need to bump to rc5 in the Github workflow file.

PR #161, opened 2023-02-27
interrupt running task if other task starts running

Ohan17:
SelectorComposite, SequenceComposite nodes also interrupt their running_child if another child return RUNNING first.

### DIFF
--- a/addons/beehave/nodes/composites/selector.gd
+++ b/addons/beehave/nodes/composites/selector.gd
@@ -27,7 +27,9 @@ func tick(actor: Node, blackboard: Blackboard) -> int:
 			FAILURE:
 				c.after_run(actor, blackboard)
 			RUNNING:
-				running_child = c
+				if c != running_child:
+					interrupt(actor, blackboard)
+					running_child = c
 				if c is ActionLeaf:
 					blackboard.set_value("running_action", c, str(actor.get_instance_id()))
 				return RUNNING
--- a/addons/beehave/nodes/composites/sequence.gd
+++ b/addons/beehave/nodes/composites/sequence.gd
@@ -29,7 +29,9 @@ func tick(actor: Node, blackboard: Blackboard) -> int:
 				c.after_run(actor, blackboard)
 				return FAILURE
 			RUNNING:
-				running_child = c
+				if c != running_child:
+					interrupt(actor, blackboard)
+					running_child = c
 				if c is ActionLeaf:
 					blackboard.set_value("running_action", c, str(actor.get_instance_id()))
 				return RUNNING
--- a/test/nodes/composites/selector_test.gd
+++ b/test/nodes/composites/selector_test.gd
@@ -44,4 +44,26 @@ func test_return_failure_of_none_is_succeeding() -> void:
 	assert_that(action1.count).is_equal(1)
 	assert_that(action2.count).is_equal(1)
 	
+func test_interrupt_second_when_first_is_succeeding() -> void:
+	action1.status = BeehaveNode.FAILURE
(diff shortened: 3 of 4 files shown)


bitbrain:
@Ohan17 can you add unit tests for this?

bitbrain:
Also, we might want to do for sequence star / selector star, as well as decorator nodes perhaps?

Ohan17:
@bitbrain I added unit tests for interrupting when success/failure and running for SelectorComposite/SequenceComposite.


> Also, we might want to do for sequence star / selector star, as well as decorator nodes perhaps?

SelectorStarComposite and SequenceStarComposite will only tick `running_child` so no other children can start running for it to be interrupted. Same for decorator nodes which have one child.
What running action should be interrupted though is the child of SelectorRandomComposite and SequenceRandomComposite. But since they're random, it will mostly likely be interrupted after 1 tick which leads me to think there should be SelectorStarRandomComposite and SequenceStarRandomComposite instead.

addons/beehave/nodes/composites/selector.gd
```diff
@@ -27,7 +27,9 @@ func tick(actor: Node, blackboard: Blackboard) -> int:
 			FAILURE:
 				c.after_run(actor, blackboard)
 			RUNNING:
-				running_child = c
+				if c != running_child:
```
bitbrain, line 30: This looks wrong to me, as we assign `running_child` to `c` prior, so it would never be different. We might need to first do this check before assigning `c` to `running_child`

bitbrain (approved):

bitbrain:
@Ohan17 this will be reverted in https://github.com/bitbrain/beehave/pull/138 as Sequence node was implemented wrongly. When a child returns RUNNING it should tick again, rather than restart. Therefore, another child would never have the opportunity to start running.

Ohan17:
@bitbrain in 2.7 shouldn't these changes be transfered to SequenceReactive and SelectorReactive since they're basically the same as previous Sequence and Selector?

bitbrain:
@Ohan17 yes, I did. Hopefully the interrupt works on them as expected.

Ohan17:
@bitbrain well the code is definitely not calling interrupt. Using the same test I wrote they currently fails.
SelectorReactive
```
func test_interrupt_second_when_first_is_running() -> void:
	action1.status = BeehaveNode.FAILURE
	action2.status = BeehaveNode.RUNNING
	assert_that(tree.tick()).is_equal(BeehaveNode.RUNNING)
	assert_that(action1.count).is_equal(1)
	assert_that(action2.count).is_equal(1)
	
	action1.status = BeehaveNode.RUNNING
	assert_that(tree.tick()).is_equal(BeehaveNode.RUNNING)
	assert_that(action1.count).is_equal(2)
	assert_that(action2.count).is_equal(0) # test fails, expecting 1
```
SequenceReactive
```
func test_interrupt_second_when_first_is_running() -> void:
	action1.status = BeehaveNode.SUCCESS
	action2.status = BeehaveNode.RUNNING
	assert_that(tree.tick()).is_equal(BeehaveNode.RUNNING)
	assert_that(action1.count).is_equal(1)
	assert_that(action2.count).is_equal(1)
	
	action1.status = BeehaveNode.RUNNING
	assert_that(tree.tick()).is_equal(BeehaveNode.RUNNING)
	assert_that(action1.count).is_equal(2)
	assert_that(action2.count).is_equal(0) # test fails, expecting 1
```

bitbrain:
@Ohan17 great catch. Addressed this in https://github.com/bitbrain/beehave/pull/166 - look out for Beehave 2.7.1

PR #186, opened 2023-04-24
Add `get_for_actor` and `set_for_actor` functions to `Blackboard`

tavurth:
## Description

`blackboard.get_value("delta", str(actor.get_instance_id()))` is often used, and takes a while to write out.

Can be nice to have a shorthand.

## Addressed issues

N/A

## Screenshots

N/A

### DIFF
--- a/addons/beehave/blackboard.gd
+++ b/addons/beehave/blackboard.gd
@@ -31,3 +31,11 @@ func has_value(key: Variant, blackboard_name: String = 'default') -> bool:
 func erase_value(key: Variant, blackboard_name: String = 'default') -> void:
 	if blackboard.has(blackboard_name):
 		blackboard[blackboard_name][key] = null
+
+
+func set_for_actor(actor: Node, key: Variant, value: Variant) -> void:
+	return self.set_value(key, value, str(actor.get_instance_id()))
+
+
+func get_for_actor(actor: Node, key: Variant, default_value: Variant) -> Variant:
+	return self.get_value(key, default_value, str(actor.get_instance_id()))


bitbrain:
What would be a usecase to get the delta from a blackboard? Also, I do not see it being set anywhere, so it will always be 0

tavurth:
You can find it here :)

https://github.com/bitbrain/beehave/blame/55fa7b1af55e9b4cdf96b3b63a5e5b9df69afaac/addons/beehave/nodes/beehave_tree.gd#L110

It's useful because on each tick we don't know for sure how much `delta` has passed. 
Changing `_physics_process` tick speed will make the game AI (effects) run faster or slower without use of delta.

tavurth:
I also added these two to my own `blackboard.gd` but I'm not sure if you would like them in this PR:

```gdscript
func set_for_actor(actor: Node, key: Variant, value: Variant) -> void:
	return self.set_value(key, value, str(actor.get_instance_id()))


func get_for_actor(actor: Node, key: Variant, default_value: Variant) -> Variant:
	return self.get_value(key, default_value, str(actor.get_instance_id()))
```

These could simplify the logic in other areas of the `Beehave` codebase however:

```gdscript
func get_delta(actor: Node) -> float:
	return self.get_for_actor(actor, "delta", 0.0)
```

Ohan17:
why would this be any different from the builtin `get_process_delta_time` and `get_physics_process_delta_time`?

tavurth:
@Ohan17 good question, I didn't ever use those.

I wonder why we're setting delta in the blackboard if those functions are available.

Oh maybe this:
https://github.com/godotengine/godot/issues/55752

bitbrain:
@tavurth great point - https://github.com/bitbrain/beehave/blob/godot-4.x/addons/beehave/nodes/beehave_tree.gd#L110 should not be needed in fact.

PR #187, opened 2023-04-25
A little nicer debugging

tavurth:
## Description

As discussed in the other PR `delta` is not needed in the `Blackboard`

I also added `microseconds` to the custom monitor, as the units were a little confusing to me before and I had to look it up.

## Addressed issues

N/A

## Screenshots

Screenshot 2023-04-25 at 18 46 32

### DIFF
--- a/addons/beehave/nodes/beehave_tree.gd
+++ b/addons/beehave/nodes/beehave_tree.gd
@@ -88,7 +88,7 @@ func _ready() -> void:
 
 	# Get the name of the parent node name for metric
 	var parent_name = actor.name
-	_process_time_metric_name = "beehave/%s-%s-process_time" % [parent_name, get_instance_id()]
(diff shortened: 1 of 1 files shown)


bitbrain:
Looks great - a little nitpick: it might be better to convert to milliseconds (not microseconds) as frame budgets are usually reasoned with in milliseconds. Ideally, Godot itself should give you the option to choose the time format but unfortunately it does not support that in the editor...

tavurth:
Hmm while true, godot expects a float or an integer.

Float cannot be rounded to a nice even number 🤔

tavurth:
Attempting to round ends up like this:

Screenshot 2023-04-25 at 21 14 25

Leaving unrounded ends up like this:
Screenshot 2023-04-25 at 21 15 05

bitbrain:
Yuck! Fine, let us merge this for now and see how we can perhaps improve Godot Engine itself to handle this better.

bitbrain (approved):

PR #190, opened 2023-05-21
Add TimeLimiter node

Ohan17:
## Description

- Add TimeLimiter node that will fails the child that does not finish running after x seconds specified in export variable `wait_time`.
- Clear running child of decorators and composite nodes after ran.
- `BeehaveRoot` now call child's `before_run` if it was not running in previous tick. 

## Screenshots

image

Player entering an Enemy range will be chased for a set amount of time, it will then lose aggro if Chase action can not finish in time.

### DIFF
--- a/addons/beehave/nodes/beehave_tree.gd
+++ b/addons/beehave/nodes/beehave_tree.gd
@@ -124,7 +124,7 @@ func _physics_process(delta: float) -> void:
 
 func tick() -> int:
 	var child := self.get_child(0)
-	if status == -1:
+	if status != RUNNING:
 		child.before_run(actor, blackboard)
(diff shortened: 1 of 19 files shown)


bitbrain (changes requested):
Fix failing test:

> Debugger Break, Reason: 'Parser Error: Could not find type "TimeLimitDecorator" in the current scope.'
*Frame 0 - res://test/nodes/decorators/time_limiter_test.gd:15 in function ''
Enter "help" for assistance.

bitbrain (approved):

bitbrain:
@Ohan17 thank you for your contribution.

PR #205, opened 2023-07-17
Replace download host with GitHub

Structed:
TuxFamily is currently down (again) and also slow.
Downloading from GitHub is faster, more reliable and therefore reduces GitHub Action runtime


bitbrain (approved):

PR #206, opened 2023-08-16
weighted random selector/sequence

### LINKED ISSUE #194: Weighted random selectors/sequences
In some use-cases (especially procedural generation), it is desired to select a random child node based on a weighted probability.  For example:

- Selector_Random
-- Child Node One (weight: 5)
-- Child Node Two (weight: 7)
-- Child Node Three (weight: 3)

Sum: 5+7+3 = 15

"Child Node One" has a (5/15) = 33% chance of selection
"Child Node Two" has a (7/15) = 47% chance of selection
"Child Node Three" has a (3/15) = 20% chance of selection

A way in which to assign a weight to a node, and have the parent composite choose one of the children based on a weighted probability.

One possible solution would be to expose a `weight` variable in `beehave_node.gd` which would be `1` by default (it would likely need to be part of this base class in order to apply to all nodes and properly support nested trees).  When a composite evaluates its children, it will sum up the weights of all children which inherit from `BeehaveNode` which, with a weight of 1, will just return the number of children and select one of them at random with equal probability.  This matches the existing behaviour of sequence_random / selector_random composites.  Changing the weights would skew the random selection according to the sum of weighted values.

**Describe alternatives you've considered**
The current functionality of Beehave allows for random selectors and sequences, but does not currently support the ability to give weights to nodes in the random selection process.

lostptr:
## Description

To better encapsulate the random manipulation logic for composites, I created a `RandomizedComposite` node that should never be used as is; it is only so that logic can be shared for `SelectorRandom` and `SequenceRandom` (or else, I would have to copy and paste a hundred lines on both scripts)

This new class contains the seed settings, option to specify weights for each child and the shuffle logic.

> ⚠️ Thorough testing needed 
> The editor for weights works fine but I noticed it is sometimes finicky. I had to reopen the scene for it to work again. Maybe this happened because I was constantly changing the code but I would like to be sure that it is all working properly.

## Addressed issues

- Closes #194 

### Weights editor:
ezgif-5-345c281904

### Proof of concept

I made a little test to make sure that the weights are actually affecting the random sampling by recording all of the times a node was picked as the first child. 
After 36698 times:
- Action "First" with weight 40 was picked 39.69% of the times
- Action "Second" with weight 30 was picked 30.03% of the times
- Action "Third" with weight 15 was picked 15.12% of the times
- Action "Forth" with weight 15 was picked 15.16% of the times
image


bitbrain:
Looking great - don't forget to also add some tests for this.

addons/beehave/nodes/composites/sequence_random.gd
```diff
@@ -3,28 +3,23 @@
 ## will be executed in a random order.
 @tool
 @icon("../../icons/sequence_random.svg")
-class_name SequenceRandomComposite extends Composite
+class_name SequenceRandomComposite extends RandomizedComposite
 
 
+signal reseted(new_order)
```
bitbrain: strange name btw, something like `state_reset` or just `reset` might be better.

lostptr: Oh dont mind that I just did that for testing purposes... (like that node_2d.tscn)
But now that you mentioned I think this will be needed for the unit tests.

lostptr: Another question: there are a couple of tests that I would need a scene (ie.: updating the weights when nodes are added in runtime) but theres only one scene test `e2e_test`. In terms of folder structure, where could I keep the new tests?

lostptr:
Github actions tests failed 😢

image
Could this be the cause?

addons/beehave/nodes/composites/sequence_random.gd
bitbrain: This looks good how you did it! Basically with the node. `e2e_test` was just something to check that the addon generally loads.

bitbrain:
@lostptr a couple of findings:

- renaming a child node affected by a weight resets the weight
- If I set the weight of Node A to 50 but the weight of node B to 10, currently node B gets invoked more often which is rather confusing - is this intentional? As your description suggestes that a higher weight equals more likelihood to be executed. e.g. if I set weight of Node A to 30 and of node B to 70, A is called way more often which seems to be wrong

[weight-issue.zip](https://github.com/bitbrain/beehave/files/12386612/weight-issue.zip)

addons/beehave/nodes/composites/randomized_composite.gd
```diff
@@ -0,0 +1,138 @@
...
+		if _weights.has(c.name):
+			new_weights[c.name] = _weights[c.name]
+		else:
+			new_weights[c.name] = 1
+	_weights = new_weights
+	notify_property_list_changed()
+
+
+func _on_child_entered_tree(node: Node):
+#	print('%s has entered the tree' % node.name)
```
lostptr: Note to self: remove this

lostptr:
> renaming a child node affected by a weight resets the weight

Yeah I noticed that too but I had no ideia how to deal with that  and thought it wasn't such I big deal.  However, I can see how that can be annoying and cause a bunch of silly bugs because people would not notice. I found a way to fix that, albeit not very elegant 😆. 

> If I set the weight of Node A to 50 but the weight of node B to 10, currently node B gets invoked more often which is rather confusing - is this intentional? As your description suggestes that a higher weight equals more likelihood to be executed. e.g. if I set weight of Node A to 30 and of node B to 70, A is called way more often which seems to be wrong

Oops!
I made a little change in the `SequenceRandomComposite` and forgot to do the same with the `SelectorRandomComposite`. The weights were having the exact opposite effect, it should work now.

bitbrain:
> > renaming a child node affected by a weight resets the weight
> 
> Yeah I noticed that too but I had no ideia how to deal with that and thought it wasn't such I big deal. However, I can see how that can be annoying and cause a bunch of silly bugs because people would not notice. I found a way to fix that, albeit not very elegant 😆.
> 
> > If I set the weight of Node A to 50 but the weight of node B to 10, currently node B gets invoked more often which is rather confusing - is this intentional? As your description suggestes that a higher weight equals more likelihood to be executed. e.g. if I set weight of Node A to 30 and of node B to 70, A is called way more often which seems to be wrong
> 
> Oops! I made a little change in the `SequenceRandomComposite` and forgot to do the same with the `SelectorRandomComposite`. The weights were having the exact opposite effect, it should work now.

@lostptr Why didn't the test fail? Or was the test broken too?

lostptr:
> Why didn't the test fail? Or was the test broken too?

There where no tests for the selector thats why. I have added them in the latest commit.

bitbrain (approved):

PR #212, opened 2023-08-29
Do not interrupt action after it finishes

### LINKED ISSUE #211: Sequence is interrupts action that has already finished.
Godot version: 4.1.1

**Describe the bug**
In a sequence node, after a long running action is finished, before the next long running action the first action is interrupted.

**To Reproduce**
Have the following tree:
image

- LongRunning1 takes 4 frames to return `SUCCESS`.
- Immediate returns `SUCCESS` immediatly.
- LongRunning takes 3 frames to return `SUCCESS`.

Even if LongRunning1 has already returned `SUCCESS`, it is still considered as "running_child" and gets interrupted before running LongRunning2.

**Expected behavior**
After LongRunning1 returns `SUCCESS`, running_child should be null and not be interrupted.

**Desktop:**
 - OS: Windows 11

**Sample Project**
[Beehave Interrupt Issue.zip](https://github.com/bitbrain/beehave/files/12444067/Beehave.Interrupt.Issue.zip)

lostptr:
## Description

Whenever a child returns `SUCCESS` or `FAILURE` we check if the child is the running_action or the running_child to set them to null. I did the same with the selector node however I could not find a case where selector would interrupt a child incorrectly (only found a case in sequence). 

## Addressed issues
Closes #211

### DIFF
--- a/addons/beehave/nodes/composites/sequence.gd
+++ b/addons/beehave/nodes/composites/sequence.gd
@@ -31,12 +31,10 @@ func tick(actor: Node, blackboard: Blackboard) -> int:
 		match response:
 			SUCCESS:
 				_cleanup_running_task(c, actor, blackboard)
-
 				successful_index += 1
 				c.after_run(actor, blackboard)
 			FAILURE:
 				_cleanup_running_task(c, actor, blackboard)
-
 				# Interrupt any child that was RUNNING before.
 				interrupt(actor, blackboard)
 				c.after_run(actor, blackboard)
--- a/addons/beehave/nodes/composites/sequence.gd
+++ b/addons/beehave/nodes/composites/sequence.gd
@@ -30,9 +30,13 @@ func tick(actor: Node, blackboard: Blackboard) -> int:
 
 		match response:
 			SUCCESS:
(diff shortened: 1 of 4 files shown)


addons/beehave/nodes/composites/sequence.gd
```diff
@@ -30,9 +30,13 @@ func tick(actor: Node, blackboard: Blackboard) -> int:
 
 		match response:
 			SUCCESS:
+				_cleanup_running_task(c, actor, blackboard)
+
```
bitbrain: remove whitespace

addons/beehave/nodes/composites/sequence.gd
```diff
@@ -30,9 +30,13 @@ func tick(actor: Node, blackboard: Blackboard) -> int:
...
 				successful_index += 1
 				c.after_run(actor, blackboard)
 			FAILURE:
+				_cleanup_running_task(c, actor, blackboard)
+
```
bitbrain: remove whitespace

bitbrain (changes requested):

bitbrain (approved):
Thank you and good job.

Issue #216, opened 2023-09-24
Add delta time in tick() function


Nrosa01:
**Is your feature request related to a problem? Please describe.**
The only way to get delta time in tick() function is action nodes is actor.get_physics_process_delta_time() which is annoying. Also user won't know at first if this is running in the physics process or the normal process.

**Describe the solution you'd like**
Add delta (time) parameter, to tick function. It can be passaed in `beehave_tree.gd#_physics_process(delta: float)` function at line 116

```gdscript
func tick(actor: Node, blackboard: Blackboard, delta: float) -> int:
	pass
```
This is also useful for condition nodes that needs to create a timer you want to manually control instead of using get_scene_tre().create_internal_tiner()...

**Describe alternatives you've considered**
I thought about a default blackboard having it but that didn't make sense

**Additional context**
It's troublesome not knowing at first how tick runs. I'll make another issue for this. I'll try to make also a PR to fix this issue tomorrow

lostptr:
No more updating the delta time in the blackboard!

Nrosa01:
Created a PR [here](https://github.com/bitbrain/beehave/pull/218)

PR #218, opened 2023-09-25
Added delta time to tick function as parameter

Nrosa01:
## Description

I made this changes to add delta time to the tick function to both the addon and the tests (delta is 0 for the tests). This way you won't need to pass delta using blackboard neither by calling actor.get_physics_process_delta_time() manually in every action or condition that requires it.

Despite all be wary of this PR, I have just tested it works in my project where I just use normal selector, normal composite, conditions and actions (without blackboard) and I haven't run the tests (I assumo a workflow will run them when I upload to github, if not pls let me know how to run them)

## Addressed issues

- Solved issue #216


bitbrain:
Thank you for this - btw, we have tests already inside the test folder. Could you please move your relevant tests [into the test classes](https://github.com/bitbrain/beehave/tree/godot-4.x/test)?

Also, this breaks compatibility as any person upgrading to this change will have errors (as they implement a tick method without delta) - this is why we require a compatibility break warning for the next 2.8 release.

Nrosa01:
Ups, I forgot about the compatibility. Maybe I could add delta as an optional parameter, so that way it doesn't break compatibility, I have yet to test if these works in gdscript as I expect.

About the tests, if I'm sincere I'm not used to test environments, I just updated the current tests so they don't fail at runtime for missing the delta parameter. Do you think I should add new tests?

bitbrain:
@Nrosa01 no need to write new tests but only adjust the existing ones. However, you did not modify them but created completely new test files. Instead, delete the newly created ones again but do the changes on the existing tests in the test folder.

No need to make it optional. It is more of a headsup for us to give a compatibility breakage warning on the release notes.

bitbrain:
@Nrosa01 tests are failing. Worth running them locally to see what breaks.

Nrosa01:
Ups sorry, I've been busier than I expected today and I couldn't check stuff properly. I'll try to see how to run the tests locally tomorrow and fix whatever is happening. I'll make sure this time everything is alright before pushing

lostptr:
Just a headsup: I think we need to change the docs for condition leaf as well and add the delta parameter.

bitbrain:
Closing this - more context here: https://github.com/bitbrain/beehave/discussions/243

Issue #220, opened 2023-09-28
interrupt doesn't get called


HoutarouOreki:
Godot version: 4.1.1
Beehave version: 2.7.5 and [cf18be7](https://github.com/bitbrain/beehave/commit/cf18be7b18d8d11f83b6dc67280d6432ba5a6abb)

**Describe the bug**
I have an ActionLeaf. 

I expect its method `interrupt` to be called when the Leaf stops being ticked while still running.

My expectations are not met.

**Scenario where it works wrong**

- GoToFetchable (1) is running
- IsFetching (2) turns to false, so GoToFetchable stops being ticked, due to (2) and (1) being under a reactive sequence
- SittingAction (3) is now being ticked

![image](https://github.com/bitbrain/beehave/assets/16261733/bff19d3d-a0f1-4974-a8e5-e1cd467eed26)

GoToFetchable (1) never gets the `interrupt` or `after_run` methods called during this scenario, both on master and 2.7.5.
By reading the manual, I thought `interrupt` would get called. I want to handle this interruption but the method doesn't get called.

To add to this, on failure/success of GoToFetchable (1):
- 2.7.5: both `after_run` and `interrupt` get called, which is wrong
- latest commit: `after_run` gets called, which is as expected

**Desktop:**
Linux Ubuntu 20.04

lostptr:
Hey, I could work on this but it would be better if we had a minimal reproduction project

HoutarouOreki:
https://github.com/HoutarouOreki/BeehaveInterruptRepro

HoutarouOreki:
Works now, thank you 💙

PR #224, opened 2023-10-01
fixed selector not calling interrupt()

lostptr:
## Description
`Selector` and `SelectorRandom` were not calling `interrupt()` because of a bug (after_run should not be called in the interrupt method)

## Addressed issues
#220

### DIFF
--- a/test/nodes/composites/selector_test.gd
+++ b/test/nodes/composites/selector_test.gd
@@ -9,6 +9,7 @@ const __source = "res://addons/beehave/nodes/composites/selector.gd"
 const __count_up_action = "res://test/actions/count_up_action.gd"
 const __blackboard = "res://addons/beehave/blackboard.gd"
 const __tree = "res://addons/beehave/nodes/beehave_tree.gd"
+const __selector_reactive = "res://addons/beehave/nodes/composites/selector_reactive.gd"
 
 var tree: BeehaveTree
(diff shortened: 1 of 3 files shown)


bitbrain:
I'd expect some unit test for this proving the bug (that should be green due to this bugfix)

test/nodes/composites/selector_test.gd
```diff
@@ -125,3 +126,29 @@ func test_not_interrupt_first_after_finished() -> void:
...
+	assert_that(tree.tick()).is_equal(BeehaveNode.RUNNING)
+	assert_that(action1.count).is_equal(1)
+	assert_that(action2.count).is_equal(0)
+	
+	fake_condition.status = BeehaveNode.SUCCESS
+	assert_that(tree.tick()).is_equal(BeehaveNode.SUCCESS)
+	assert_that(action1.count).is_equal(0)
+	assert_that(action2.count).is_equal(0)
+	
+	# clean up...
```
bitbrain, line 151: generally, individual tests should not do the cleanup but the clean up should happen in an "after_test" step

lostptr: yes but after_test would be called for every test in the file and this clean up should only happen in this particular test

bitbrain (approved):

Issue #219, opened 2023-09-25
Debugger crashes Godot 4.2-dev5


RafaelVidaurre:
Godot version: 4.2-dev5

**Describe the bug**
Opening the debugger graph edit view crashes the engine (not a standard error, a crash).

As far as I could tell, the engine crashes when a node is added as a child of the GraphEdit node.

**To Reproduce**
1. Run the game with at least one active behavior tree
2. Go to the debugger and click on the Beehive tab
3. Click on any behavior tree from the list

**Expected behavior**
The graph showing the behavior tree should appear.

**Screenshots**

https://github.com/bitbrain/beehave/assets/1808674/8bbc3aa5-3c2d-450d-b5c3-44fc44baefd7

**Desktop (please complete the following information):**
 - OS: MacOS, Sonoma

**Additional context**
Updating the function names of `get_zoom_hbox()` does not solve this, although this change is also required due to GraphEdit's latest breaking changes.

P.S. Even though there's an incomplete PR #214  attempting to solve this issue, I thought it'd be better to have this as a proper bug report

bitbrain:
This will be resolved by https://github.com/bitbrain/beehave/pull/214

PR #226, opened 2023-10-04
Fix godot 4.2-dev6 compatibility issues

RafaelVidaurre:
## Description

This PR deals with the breaking changes introduced in 4.2 on the `GraphEdit` API in a way that is backwards compatible.

* I'm seeing some errors regarding GDUnit which I'm guessing will get resolved once GDUnit supports 4.2.
* Introduces the **minimum** required changes for Beehave debugger to be able to run on 4.2-dev6
* Changes the example scene to use `SelectorReactiveComposite` since `SelectorStarComposite` doesn't exist anymore
* Does not address rendering issues with the debugger (see attached screenshot) 

## Addressed issues

- Editor crash when opening the debugger
- Fix issue with reference to non-existent global class `SelectorStarComposite`

## Screenshots
Screenshot 2023-10-04 at 11 11 56

Solves #219

### DIFF
--- a/addons/beehave/debug/debugger_tab.gd
+++ b/addons/beehave/debug/debugger_tab.gd
@@ -41,7 +41,7 @@ func _ready() -> void:
 	button.pressed.connect(func(): make_floating.emit())
 	button.tooltip_text = "Make floating"
 	button.focus_mode = Control.FOCUS_NONE
-	graph.get_zoom_hbox().add_child(button)
+	graph.get_menu_container().add_child(button)
 
 	var toggle_button := Button.new()
 	toggle_button.flat = true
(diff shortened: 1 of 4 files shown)


bitbrain:
Screenshot 2023-10-09 083608

Looks good to me! I noticed while testing with 4.1.stable that the debug tree does not appear. Can anybody else reproduce this?

@RafaelVidaurre in your screenshot, the outputs of each box seems to be aligned wrongly, too. This is how it is supposed to look like:

debug

I am happy to create a separate bug ticket for it to fix the visual fidelity of the debugger separate to this PR for 4.2 in case this turns out to be too difficult to do.

RafaelVidaurre:
I think creating a separate ticket would be ideal, so that we can get Beehive to at least run on v4.2

bitbrain:
@RafaelVidaurre makes sense - however, currently this PR breaks the debugger in 4.1 (which most people still use right now), could you have a look why in 4.1 the debugger does not show at all? Perhaps it is just me (see screenshot above)

bitbrain:
I solved the additional Beehave issues now in #231 and will close this PR.

PR #235, opened 2023-10-14
repeater decorator + tests

### LINKED ISSUE #230: Repeater Decorator
### **Is your feature request related to a problem? Please describe.**
I would like to suggest a decorator to repeat their child X times as I have many enemies that would use this behavior and I dont think is too specific to my project.

### **Describe the solution you'd like**
A decorator that runs its child for the specified number of times or when the child returns `FAILURE`. A child execution is only counted when it returns `SUCCESS` (if returns `FAILURE` it stops immediatly, returning `FAILURE` as well).

Also, this could be the icon: (keep in mind that the Limiter's icon is this image)
image

### **Describe alternatives you've considered**
I cannot use a Limiter to this purpose for two reasons. 
It doesnt work as I expect with long running tasks and it always returns FAILURE, forcing me to always use it with a Succeeder.

To quote the docs:
> The `Limiter` node executes its child a specified number of times (x). When the maximum number of ticks is reached, it returns a `FAILURE` status code. 

This is good to use in conditions that return imediately but it doesnt work well with long running tasks. It could be a bug but right now if a Limiter is set to execute 3 times it will do this:
```
Limiter::before_run()
child.tick() -> RUNNING, returning RUNNING as well
child.tick() -> RUNNING, returning RUNNING as well
child.tick() -> RUNNING, returning RUNNING as well
done executing child, returning FAILURE
Limiter::after_run()
```

lostptr:
## Description

Adds the repeater decorator along with tests.

## Addressed issues

Closes #230


bitbrain (approved):
Nice, LGTM!

PR #236, opened 2023-10-18
Move Class Descriptions Below Class Names

mechPenSketch:
## Description

Class descriptions are now placed under `class_name` declarations. This allows the descriptions to be seen in the editor.

[GDScript documentation comments](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_documentation_comments.html)

Property descriptions will also appear in the Inspector.

## Addressed issues

- l did not post any issues on this.

## Screenshots

godot_desc
godot_desc2
godot_desc3


bitbrain (approved):
Given we want to make this addon compatible across most minor 4.x versions, and this particular issue had only been fixed in 4.2 I am happy with this change.

PR #239, opened 2023-10-25
Add Moving Script Templates to Installation

mechPenSketch:
## Description
On `README.md`, I have added one more step to the Installation section. This lets users know that by moving folder `script_templates` into their projects, they can access the script templates.

For consistency, I have also changed folder `beehave` into `addons/beehave` so that users would know that `beehave` is in `addons`.

## Screenshots
godot_script_template2
godot_script_template


bitbrain (approved):
Nice!

PR #249, opened 2023-11-16
🐛 fix interrupt of sequence star/random + improve docs

### LINKED ISSUE #248: Documentation for the behaviour of composite nodes is misleading
The current documentation for the behaviour of composite nodes contains conflicting information and inaccuracies.

Firstly, both Sequence and Selector documentation pages state the following:

> Every tick, the Selector/Sequence node processes all its children, even if one of them is currently RUNNING.

This is wrong, since only the Reactive variants process all child nodes, regardless of the return status.

Secondly, the **Types of Sequence/Selector Nodes** tables in the Sequence and Selector documentation pages correctly illustrate the tick behaviour of each composite node sub-type, depending on the child node's return status. However, it uses the term _restart_ to mean that the composite starts ticking from the first child, as opposed to _tick again_ which means that it ticks from where it left off the last time. This is confusing, because the term 'restart' is explained in [Composites](https://bitbra.in/beehave/#/manual/composites?id=restarting-a-composite) page as follows:

> **Restarting a composite**
>
> If a parent node restarts a child node, it means that the parent node will start the child node from scratch the next time it is ticked. This means that any progress made by the child node will be reset, and it will start its execution from the beginning.

This suggests it's the child node that gets restarted - i.e. it receives the `before_run` callback - and not the execution of the sequencer node itself gets restarted. This is contrary to how the term is used in the tables on Sequence and Selector pages. It is also contrary to the heading, i.e. *Restarting a composite** suggests it's the composite itself that gets restarted, not the child node.

Relevant documentation pages:
https://bitbra.in/beehave/#/manual/composites
https://bitbra.in/beehave/#/manual/selector
https://bitbra.in/beehave/#/manual/sequence

bitbrain:
## Description

Usually, any sequence node that fails should correctly interrupt its running children, if there are any. This worked correctly for `sequence` and `reactive sequence` but not for sequence star nor sequence random.

Also addressed docs, which hopefully closes #248

### DIFF
--- a/docs/manual/composites.md
+++ b/docs/manual/composites.md
@@ -18,12 +18,15 @@ With composite nodes, you can craft unique and dynamic Behavior Trees for your g
 
 ## Restarting a composite
 
-When a parent node restarts, it means the whole composite node begins its evaluation again from the beginning. This does not interrupt the child nodes.
-
-## Interrupting child nodes
-
-A sequence may interrupt any `RUNNING` child node in case a child returns a `FAILURE`. The `interrupt` method will be called on all child nodes of the currently evaluated child node.
+When a parent node restarts, it means the whole composite node begins its evaluation again from the beginning. This does not interrupt the running child node.
 
 ## Ticking again a composite
 
-If a parent node ticks a child node again, it means that the parent node will immediately tick the child node again on the next frame, without waiting for the current frame to finish executing. This allows the child node to continue its execution from where it left off without resetting its progress.
+When a composite node ticks again, it will "jump" to the currently `RUNNING` child node and tick it.
+
+For [sequence nodes](sequence.md), on the first frame it starts ticking all its children in order, starting from the first one, until either all of them return `SUCCESS` or one of them returns `RUNNING`. If one of them returned `RUNNING`, on the next frame, the sequence will only tick the running child node. Not all of them, just the running one - and possibly the child nodes following it, if the running child has completed and returned `SUCCESS`. 
+For [selector nodes](selector.md), on the first frame it starts ticking each child in order until one of them returns `SUCCESS` or one of them returns `RUNNING`. If one of them returned `RUNNING`, on the next frame, the selector will only tick the running child node, skipping all the previous ones, and possibly the child nodes following it, if the running child has completed and returned `FAILURE`. 
+
+## Interrupting child nodes
+
+A sequence may interrupt any `RUNNING` child node. The `interrupt` method will be called recursively on all descendant nodes of the interrupted child node.
--- a/docs/manual/composites.md
+++ b/docs/manual/composites.md
@@ -18,10 +18,12 @@ With composite nodes, you can craft unique and dynamic Behavior Trees for your g
 
 ## Restarting a composite
 
-If a parent node restarts a child node, it means that the parent node will start the child node from scratch the next time it is ticked. This means that any progress made by the child node will be reset, and it will start its execution from the beginning.
+When a parent node restarts, it means the whole composite node begins its evaluation again from the beginning. This does not interrupt the child nodes.
+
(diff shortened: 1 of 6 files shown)


docs/manual/composites.md
```diff
@@ -18,10 +18,12 @@ With composite nodes, you can craft unique and dynamic Behavior Trees for your g
 
 ## Restarting a composite
 
-If a parent node restarts a child node, it means that the parent node will start the child node from scratch the next time it is ticked. This means that any progress made by the child node will be reset, and it will start its execution from the beginning.
+When a parent node restarts, it means the whole composite node begins its evaluation again from the beginning. This does not interrupt the child nodes.
```
rxlecky: If I'm not mistaken, there can only be one running node - that is until parallel composites are implemented. But until then, I think the docs should reflect this fact and be clear about it as not to confuse the new users. Therefore, I suggest the last sentence be rewritten as follows: `This does not interrupt the running child node.`

docs/manual/composites.md
```diff
@@ -18,10 +18,12 @@ With composite nodes, you can craft unique and dynamic Behavior Trees for your g
...
 
-## Ticking again a composite
+## Interrupting child nodes
+
+A sequence may interrupt any `RUNNING` child node in case a child returns a `FAILURE`. The `interrupt` method will be called on all child nodes of the currently evaluated child node.
```
rxlecky: > A  sequence may interrupt any `RUNNING` child node in case a child returns a `FAILURE`.

The interrupt can also happen in other cases, not just when a child returns a failure - for example, if a reactive sequence encounters a running child, that comes before the child that was running last frame, the old running child is interrupted. I'd say this part of the sentence can be left out: `in case a child returns a FAILURE.`

rxlecky: > The `interrupt` method will be called on all child nodes of the currently evaluated child node.

The term _child nodes_ typically denotes only the direct children, i.e. the first-order descendants. For clarity, I'd suggest using the term _descendant nodes_ to denote all descending nodes and say `... will be called on all descendant nodes`.

Also, instead of `the currently evaluated child node`, I'd say `the interrupted child node` just to make it clear which node we are referring to.

docs/manual/composites.md
```diff
@@ -18,10 +18,12 @@ With composite nodes, you can craft unique and dynamic Behavior Trees for your g
...
 
-If a parent node ticks a child node again, it means that the parent node will immediately tick the child node again on the next frame, without waiting for the current frame to finish executing. This allows the child node to continue its execution from where it left off without resetting its progress.
+## Ticking again a composite
 
-In other words, restarting a child node means that the parent node will give the child node a fresh start, while ticking it again means that the parent node will let the child node continue its execution from where it left off.
+If a parent node ticks a child node again, it means that the parent node will immediately tick the child node again on the next frame, without waiting for the current frame to finish executing. This allows the child node to continue its execution from where it left off without resetting its progress.
```
rxlecky: >..., without waiting for the current frame to finish executing.

Is this correct? This sounds like the child node is ticked multiple times in the same frame. Isn't each node only ticked at most once on every frame?

bitbrain: Brain fog on my part. I will correct this.

rxlecky: If I may add one more thing, I'd also maybe elaborate more to clarify that all that ticking again really means is that the parent composite node carries on with its execution from that given child node. Meaning, that it's not only that given child node that is ticked, it is potentially its sibling nodes following it, if the node completes and allows its siblings to be ticked too.

bitbrain:
@rxlecky incorporated your feedback here: https://github.com/bitbrain/beehave/pull/249/commits/b07b02d8f6c3c522e4cbb25ffbf2546fbf136e7d

PR #250, opened 2023-11-16
:bug: fix limiter node behaviour

### LINKED ISSUE #247: Limiter and TimeLimiter exhibit inconsistent behaviour
Godot version: 4.1.2
Beehave version: 2.7.6

**Describe the bug**
Limiter and TimeLimiter decorator nodes exhibit inconsistent behaviour, compared to other nodes. Here's a list of inconsistencies I've noticed:

1. Limiter and TimeLimiter only call the `after_run` callback on their child node if they interrupt it. If the child completes on its own, `after_run` callback doesn't get called.
2. Limiter calls only the `after_run` callback on its child node after it interrupts it. Calling `interrupt` just before `after_run` is the expected behaviour.
3. When interrupting its child node, TimeLimiter calls the `after_run` callback before the `interrupt` one. Calling `interrupt` before `after_run` is the expected behaviour.
4. Limiter never resets its counter when the node runs anew. This is inconsistent with TimeLimiter which resets every time it's run again after it completes. The documentation doesn't mention resetting, so this may actually be the expected behaviour but it should at least be documented accordingly. Or maybe a toggle can be added to make it behave one way or another, depending on the user's need, as I can see a use case for both behaviours.

I lumped a bunch of issues together into one since they are all (somewhat) related. I'm happy to split things up if necessary.

bitbrain:
Closes #247

### DIFF
--- a/addons/beehave/nodes/beehave_tree.gd
+++ b/addons/beehave/nodes/beehave_tree.gd
@@ -124,6 +124,8 @@ func _physics_process(delta: float) -> void:
 
 
 func tick() -> int:
+	if actor == null:
+		return FAILURE
 	var child := self.get_child(0)
 	if status != RUNNING:
 		child.before_run(actor, blackboard)
@@ -143,6 +145,9 @@ func tick() -> int:
 
 func _get_configuration_warnings() -> PackedStringArray:
 	var warnings:PackedStringArray = []
+	
+	if actor == null:
+		warnings.append("Configure target node on tree")
 
 	if get_children().any(func(x): return not (x is BeehaveNode)):
 		warnings.append("All children of this node should inherit from BeehaveNode class.")
--- a/addons/beehave/nodes/beehave_tree.gd
+++ b/addons/beehave/nodes/beehave_tree.gd
@@ -74,11 +74,6 @@ func _ready() -> void:
 	if Engine.is_editor_hint():
 		return
 
-	if self.get_child_count() > 0 and not self.get_child(0) is BeehaveNode:
-		push_warning("Beehave error: Root %s should have only one child of type BeehaveNode (NodePath: %s)" % [self.name, self.get_path()])
-		disable()
-		return
-
 	if not blackboard:
 		_internal_blackboard = Blackboard.new()
 		add_child(_internal_blackboard, false, Node.INTERNAL_MODE_BACK)
--- a/addons/beehave/nodes/beehave_tree.gd
+++ b/addons/beehave/nodes/beehave_tree.gd
@@ -124,7 +124,7 @@ func _physics_process(delta: float) -> void:
 
 
 func tick() -> int:
-	if actor == null:
+	if actor == null or get_child_count() == 0:
 		return FAILURE
 	var child := self.get_child(0)
 	if status != RUNNING:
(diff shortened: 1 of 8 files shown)


addons/beehave/nodes/beehave_tree.gd
```diff
@@ -124,6 +124,8 @@ func _physics_process(delta: float) -> void:
 
 
 func tick() -> int:
+	if actor == null:
```
bitbrain: there was a scenario where you can set a `null` actor on a behaviour tree. In this case, the tree should not break the game but show a configuration warning.

addons/beehave/nodes/decorators/limiter.gd
```diff
@@ -10,7 +10,10 @@ class_name LimiterDecorator extends Decorator
 @export var max_count : float = 0
 
 func tick(actor: Node, blackboard: Blackboard) -> int:
-	var child = self.get_child(0)
+	if not get_child_count():
```
bitbrain: if for whatever reason there is no child node added to this node, we should skip this behaviour (and show a configuration warning)

addons/beehave/nodes/decorators/time_limiter.gd
```diff
@@ -8,14 +8,22 @@ class_name TimeLimiterDecorator extends Decorator
...
 @export var wait_time: = 0.0
 
-var time_left: = 0.0
-
-@onready var child: BeehaveNode = get_child(0)
+@onready var cache_key = 'time_limiter_%s' % self.get_instance_id()
 
 
 func tick(actor: Node, blackboard: Blackboard) -> int:
+	if not get_child_count():
```
bitbrain: if for whatever reason there is no child node added to this node, we should skip this behaviour (and show a configuration warning)

addons/beehave/nodes/decorators/time_limiter.gd
```diff
@@ -28,20 +36,22 @@ func tick(actor: Node, blackboard: Blackboard) -> int:
...
+			child.after_run(actor, blackboard)
 		return response
 	else:
-		child.after_run(actor, blackboard)
 		interrupt(actor, blackboard)
+		child.after_run(actor, blackboard)
 		return FAILURE
 
 
-func before_run(actor: Node, blackboard: Blackboard) -> void:
```
bitbrain, line 39: not required, we will do it once when time left == 0.0 (see line 22)

rxlecky: This is (possibly) a breaking change, because originally the timer was reset when the limiter received `before_run` callback but with the current changes, it is never reset. This resetting behaviour is something I'm currently relying on in my behaviour tree.

bitbrain: Good point. Let me ensure this behaviour is consistent - I'll revert this.

addons/beehave/nodes/composites/composite.gd
```diff
@@ -7,14 +7,6 @@ class_name Composite extends BeehaveNode
 var running_child: BeehaveNode = null
 
 
-func _ready():
-	if Engine.is_editor_hint():
-		return
-
-	if self.get_child_count() < 1:
-		push_warning("BehaviorTree Error: Composite %s should have at least one child (NodePath: %s)" % [self.name, self.get_path()])
```
bitbrain, line 15: `_get_configuration_warnings` should be sufficient. No need to spam the console

bitbrain:
@rxlecky I have adjusted the PR - it was previously not clear when exactly a `TimeLimiter` and `Limiter` would work. I clarified in the docs that it only works with children that return `RUNNING` and in other cases it resets.

PR #276, opened 2023-12-23
Delay and Cooldown Decorators

RedstoneParadox:
## Description

Adds a new delay decorator, which allows nodes to be executed after a predetermined delay, which was discussed in #242. ~~Something also discussed was some way of implementing cooldowns for certain actions (i.e. a boss does an action but can't do it again for 10 seconds). My first thought was to add a mode to the delay decorator where it returns `FAILURE` instead of running while the timer is running, but I wanted to see if it would be preferable to make a separate node.~~ Edit: Decided to implement it as a separate node.

## Addressed issues

Addresses #242


RedstoneParadox:
So it would seem that `test_return_success_after_delay` doesn't always succeed; not entirely sure why that is. Edit: It's possibly a timing issue related to the use of the physics engine delta time, which I did to try and ensure that the next tick after the runner simulation would always be the first tick where the timer was 0.

Edit 2: I Rewrote the test and now it passes consistently; the test was badly written originally.

Emma-Black:
This is amazing, thank you!

bitbrain:
@RedstoneParadox can you please also add docs to the wiki about these new nodes?

RedstoneParadox:
Docs added.

bitbrain (approved):
Amazing work, thank you!

PR #287, opened 2024-01-06
🚧 [WIP] reimplement all nodes in C++

bitbrain:
## Description

This lays the foundation for **Beehave 3.0** - more details can be found here: https://github.com/bitbrain/beehave/discussions/285

### Functionalities

The following first need consultation with @bitbrain, as with Beehave V3 the idea is to design interrupts & before/after signals differently to V2.

- [ ] interrupts
- [ ] before/after signals
- [ ] time-strategies (delta vs total)

### Base Nodes

- [x] implement & test `BeehaveTree`
- [ ] implement & test `BeehaveBlackboard`

### Decorators

- [x] implement & test `BeehaveCooldown`
- [x] implement & test `BeehaveDelayer`
- [x] implement & test `BeehaveFailer`
- [x] implement & test `BeehaveSucceeder`
- [x] implement & test `BeehaveInverter`
- [x] implement & test `BeehaveLimiter`
- [x] implement & test `BeehaveUntilFail`
- [x] implement & test `BeehaveTimeLimiter`

### Composites

- [ ] implement & test `BeehaveSelector`
- [ ] implement & test `BeehaveSelectorRandom`
- [ ] implement & test `BeehaveSelectorReactive`
- [ ] implement & test `BeehaveSequence`
- [ ] implement & test `BeehaveSequenceRandom`
- [ ] implement & test `BeehaveSequenceReactive`
- [ ] implement & test `BeehaveSequenceStar`

### Leafs

- [x] implement & test `BeehaveAction`
- [ ] implement & test `BeehaveCondition`


extension/src/nodes/node.cpp
```diff
@@ -0,0 +1,20 @@
+#include "node.h"
```
zvolchak: I'd recommend avoid using "node", since it is already taken by Godot. 
Also, if you'd like to follow [C++ naming convention](https://manual.gromacs.org/5.1.1/dev-manual/naming.html#files), then I'd recommend using the same file name as the class name - `BeehaveNode.ccp` and `BeehaveNode.h` in this case.

extension/src/nodes/blackboard.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+BeehaveBlackboard::~BeehaveBlackboard()
+{
+    
+}
+
+void BeehaveBlackboard::_bind_methods()
+{
+
+}
```
zvolchak: nit: missing new line at the end of file. This is usually not a big deal, but might cause some [inconsistent or unexpected behavior in some cases](https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file). Plus, github would always warn you about it, as it does here.

P.S. Most popular IDEs these days has an option to auto-add end-of-file-new-line on Save of file close - so you never would need to think about this again.

RafaelVidaurre:
Curious, what's the current state of the V3 migration?

bitbrain:
@RafaelVidaurre see the checkboxes in the description for the current status. In terms of work, I aim to do a bit here and there, but it can take a while because I do this in my freetime as a hobby. ♥️

bitbrain:
PSA: I don't have time or resources right now to complete this migration at the moment. For anyone brave enough, I am willing to review the PRs related to this just ping me.

LenNerd42:
@bitbrain you mentioned you have some ideas for designing interrupts and before/after signals differently. Since I'm currently re-implementing some nodes that had interrupt and start/end logic - do you still remember what your design ideas were and can I find them written down anywhere?

bitbrain:
@LenNerd42 for now, adopt 1:1 what we did in gdscript as well. I initially had a different idea but it is too complicated as a first version.

Issue #240, opened 2023-11-01
Add Until Fail Decorator


meticoeus:
**Is your feature request related to a problem? Please describe.**
I would like to suggest adding the repeat Until Fail decorator. The decorator would run its child, then repeat it if it returned a success and continue to repeat it until it returns a failure.

**Describe the solution you'd like**
Add a Decorator UntilFail that acts similar to the Repeater node except it always repeats its child node when the child node returns a success, stopping when the child returns a failure, at which point the UntilFail node will return success to its parent.

**Describe alternatives you've considered**
I've looked at the Repeater but it stops after a predetermined number of successes while the desired behavior of this node is to repeat indefinitely until the child node returns a failure.

**Additional context**
Motivation: implementing an indeterminate number of repeats based on some condition or set of conditions. [This article](https://www.gamedeveloper.com/programming/behavior-trees-for-ai-how-they-work) explains a use-case involving poping data from a stack in the blackboard and going through a subtree based on that data until the stack is empty. I plan to do something similar in a project I'm working on. The number of repeats may grow while the child is repeating long running tasks due to an external actor adding more entities to the stack the sub tree will be using to loop.

PR #296, opened 2024-01-09
Until Fail Decorator

RedstoneParadox:
## Description

Adds a new decorator that runs its child and repeats until the child returns `FAILURE`, at which point it returns `SUCCESS`.

The original suggestion by @meticoeus provides additional context:

> Motivation: implementing an indeterminate number of repeats based on some condition or set of conditions. [This article](https://www.gamedeveloper.com/programming/behavior-trees-for-ai-how-they-work) explains a use-case involving poping data from a stack in the blackboard and going through a subtree based on that data until the stack is empty. I plan to do something similar in a project I'm working on. The number of repeats may grow while the child is repeating long running tasks due to an external actor adding more entities to the stack the sub tree will be using to loop.

## Addressed issues

Addresses #240


RedstoneParadox:
I'm not sure why the unit tests failed; I ran them personally and they all passed, including the new one. Edit: I forgot to update it after changing the decorator to return success instead of failure.

bitbrain (approved):

ryan-linehan:
Can we close out https://github.com/bitbrain/beehave/issues/240 now that this has been merged?

PR #307, opened 2024-02-07
Dont override actor in _ready if already defined

LunarTides:
## Description

If you set the actor in `BeehaveTree` before `_ready` is executed, it would override the actor that you set.

Now it only sets the actor if it isn't already set.

## Addressed issues

- None

## Screenshots

N/A


bitbrain:
Could you add a test for this? Happy to help in case you don't know how.

LunarTides:
Done

bitbrain:

bitbrain (approved):

PR #310, opened 2024-02-15
Fix reloading scene debug issue

zeroregard:
## Description

When you reload the scene, it seems that behaviour trees are not unregistered properly for debugging purposes. That means they will keep being in the Beehave debug tab and the Debugger output will be flooded with a lot of error messages.

## Screenshots

image

### DIFF
--- a/addons/beehave/debug/global_debugger.gd
+++ b/addons/beehave/debug/global_debugger.gd
@@ -13,7 +13,7 @@ func _on_debug_message(message: String, data: Array) -> bool:
 		_set_active_tree(data[0])
 		return true
 	if message == "visibility_changed":
-		if _active_tree:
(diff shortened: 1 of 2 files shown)


addons/beehave/debug/global_debugger.gd
```diff
@@ -13,7 +13,7 @@ func _on_debug_message(message: String, data: Array) -> bool:
 		_set_active_tree(data[0])
 		return true
 	if message == "visibility_changed":
-		if _active_tree:
+		if _active_tree && is_instance_valid(_active_tree):
```
zeroregard, line 16: Also had to add this because the previously active tree gets chopped when reloading the scene.

bitbrain:
Amazing, thanks!

bitbrain (approved):

PR #318, opened 2024-02-18
Aadded max wait time option to the time_limiter node

alextkd2003:
## Description

There were no problems, but I think it would be useful when there are several instances of the same npc using this node, since we can get some random behavior.

## Addressed issues

N/A

## Screenshots

time limiter max wait time

### DIFF
--- a/addons/beehave/nodes/decorators/time_limiter.gd
+++ b/addons/beehave/nodes/decorators/time_limiter.gd
@@ -7,6 +7,8 @@ class_name TimeLimiterDecorator extends Decorator
 ## The timer resets the next time that a child is not `RUNNING`
 
 @export var wait_time: = 0.0
+@export var max_wait_time: = 0.0
+var current_wait_time: = 0.0
 
 @onready var cache_key = 'time_limiter_%s' % self.get_instance_id()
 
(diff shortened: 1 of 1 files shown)


bitbrain:
Can you elaborate further what the issue is you are trying to solve with this?

Also, what exactly is the difference between wait time and max wait time?

alextkd2003:
hi @bitbrain this is not an issue is a nice to have feature. I constantly use the time_limiter node. I modified mine just to have that feature.

Let's say one npc is in idle mode and I'm using the time_limiter to set it to stay idle for 5 seconds and then go for a walk. If all my copies have the same value then they will all walk at the same time. If I change the values for copies, there will always be a pattern. But if the time is random, then I will have less predictable behavior since the states will be changing. In this way and all the copies of the npc will be doing different things.

I will add the same feature to all nodes with a timer to be honest, this is fantastic!! imo. I used in the pass Behavior Designer for Unity and I found that feature really cool, again imo.


Tree Ex.
image

alextkd2003:
About the differences about **wait_time** and **max_wait_time**. If max_wait_time is specified wait_time is the min_wait_time, if not specified and remains at 0 it works as it has always been working.

It may not be the best way to call the variables, but I wanted to change very little. I'll let you evaluate if these is a good idea, if not, nothing happens, in the same way all the scripts can be edited so the user can make this change if need it.

Thank you for all the effort you have put into this addon, of all the BTs I have tried there is no doubt that this is the best.

bitbrain:
After much consideration, this is likely not going into Beehave V2, given its very specific and custom use case. However, I still think this is an interesting feature so [feel free to raise a discussion here](https://github.com/bitbrain/beehave/discussions/categories/ideas).

Issue #304, opened 2024-01-30
Add tree setup examples to the wiki


tristan-f-r:
Currently, the only part of the docs that covers how to structure your tree is the videos and a single mention of them in the introduction. The docs, after "Actions and Conditions," as well as in "Buiding your Behavior Tree," should have example tree structures and their usages, possibly something even interactive (to see the individual scripts in nodes), if the current docs software allows that. This can also possibly prevent issues like #303.

I would be interested in working on this if it's approved.

lostptr:
We do lack an updated video but with 3.0 coming and don't know if things are going to change yet again.

PR #354, opened 2024-09-18
Some tree views for examples

suhankins:
## Description

First step towards solving #304. Adds tree views for some examples.
Unfortunately, not all. Some examples are too vague to really be illustrated as tree views, and some are too complicated.

## Addressed issues

#304

## Screenshots

image
image
image
image
image


bitbrain (approved):
Wow, this is a great start. Good job!

PR #358, opened 2024-10-15
Optional time slicing system

ghost:
## Rational
This pull request is a proof of concept for a opt-in time slicing system.  This is useful for games with large amounts of entities using beehave trees at once or simply ensuring a fixed amount of updates happen each frame on lower end systems such as mobile.  Both these cases aren't covered by changing the tick rate or can be solved better through this implementation.  Tick rate alone forces trees to update slower even when there is not any load and leaves the potential for excessive updates to happen on an individual frame.

This should be a part of beehave rather than left up to the user because it requires directly modifying the tick method.

## Changes
- The added global autoload executes a fixed amount of ticks per frame.
- The added export variable to ```beehave_tree.gd``` opts a tree into the time slicing system.
- The changes to ```plugin.gd``` are just to clean up the plugin's autoloads when the plugin is disabled.

### DIFF
--- a/addons/beehave/nodes/beehave_tree.gd
+++ b/addons/beehave/nodes/beehave_tree.gd
@@ -30,6 +30,11 @@ signal tree_disabled
 ## tick() runs every frame.
 @export var tick_rate: int = 1
 
+## If the tree should opt into time slicing.  If enabled, the tree will not
+## execute it's own ticks directly.  It will instead push its ticks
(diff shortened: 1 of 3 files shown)


bitbrain:
Beehave now supports `MANUAL` process mode, so a feature like this can be implemented outside of Beehave easily without the need to bake it into the addon directly.

PR #365, opened 2025-01-17
Use constants where possible

Flynsarmy:
## Description

Magic numbers are bad - if the enum list ever changes we could have unintended problems.


bitbrain (approved):
Amazing PR, thanks!

PR #366, opened 2025-01-17
Fade debug nodes back to normal state

Flynsarmy:
## Description

Fades debug nodes from SUCCESS/FAILURE back to normal instead of blinking them.

## Addressed issues

- When leaf nodes complete in 1 tick they blink a SUCCESS/FAILURE color in the debugger for just that frame and it can often be hard to spot which makes debugging difficult

## Screenshots

[show your changes with screenshots](https://github.com/user-attachments/assets/76aea575-49c2-460d-b006-3bfd039bc277)


bitbrain (approved):
Love this new addition! Thank you.

PR #368, opened 2025-01-19
Add blackboard debugging

Flynsarmy:
## Description

Adds debugging of blackboard data. Select one or more nodes in the debug tab and the blackboard data associated with that node appears on the right. The data updates in real time. RIght column is resizeable.

This PR is backwards compatible as the new argument added to `BeehaveDebuggerMessages.process_` isn't required and defaults to an empty dictionary.

## Addressed issues

- It can be difficult to know what your blackboard is doing a state machine is running. This PR greatly helps with that.

## Screenshots

https://github.com/user-attachments/assets/8cc0db58-b5ea-4553-b40a-07791d6b785d


bitbrain:
This is insanely useful. Thank you for this! 🌟

bitbrain (approved):

PR #376, opened 2025-02-28
Cooldown should save running_child so that its non-ActionLeaf childre…

tangx246:
Currently, cooldown.gd does not properly save running_child if it is not an ActionLeaf. This causes interrupt() to not propagate down the tree


bitbrain:
@tangx246 can you add a unit test to verify this works?

tangx246:
> @tangx246 can you add a unit test to verify this works?

I have done this. I'm a bit unfamiliar with the framework, so let me know if I screwed anything up. I have confirmed running these tests against the old code that the fix does what it's supposed to do

armedturret:
I noticed this bug applies to the Delayer and Limiter decorators as well! Seems to be an identical fix.

bitbrain:
@armedturret feel free to raise another PR for this. I'll merge this one.

bitbrain (approved):

PR #384, opened 2025-03-17
feat(inline-debug): add visual runtime debug node

adadgio:
## Description

I found particularily useful to have a debug node easily usable inside a scene on top of the debugger to check objects current action status. This PR addresses this feature. 

## Usage

Slap the node anywhere in your scene, attach a tree (as exported property), and see the current sequence/action/status at runtime.

## Addressed issues

- More debug at runtime

## Screenshots

Screenshot_2025-03-17_at_21_55_18
Screenshot 2025-03-17 at 21 53 56
Screenshot 2025-03-17 at 21 54 20

## Notes

Please comment on this idea to make it more modular. At this stage, you can add as many PER tree, and it only shows the current sequence / action / status triple. Its movable on the stage as any Godot node.


adadgio:
This is too complicated and there are other easier ways to to this, like this for example

```
class_name BeehaveInlineDebug
extends PanelContainer

@export var tree: BeehaveTree
@onready var action_name: Label = %ActionName

var current_action

func _ready() -> void:
	pass


func _process(_delta: float) -> void:
	if tree:
		current_action = tree.get_running_action()
		update_info()


func update_info() -> void:
	if current_action:
		action_name.text = "%s" % current_action.name
	else:
		action_name.text = ""

```

Issue #340, opened 2024-06-02
Cooldown should reset automatically when moving to another branch


Nodragem:
**Is your feature request related to a problem? Please describe.**
When implementing an attack cooldown, we want the enemy to attack immediately the first time, then wait for the cooldown.
However, at the moment, if the selector moves to another behavior the cooldown is not resetted. Thus, when we come back to the Attack sequence, we need to wait for the cooldown to be over for the attack to happen.

See video, the first time the IA goes in `GoToAttack`, the cooldown works as intended. Then I move away, the IA goes to `GoToReach`. It reaches me and come back to `GoToAttack`. Then, as the cooldown was never resetted, the IA has to wait the cooldown to attack.

https://github.com/bitbrain/beehave/assets/10520249/a4c3c551-b9e4-4829-8fd5-493ee8c13c79

Note: for the purpose of the video I disable the action `TryStopAttackCooldown`.

**Describe the solution you'd like**
Somehow, when the cooldown is interrupted, it should be resetted. 

**Describe alternatives you've considered**
I played with the `interrupted` and `run_after`, but as the cooldown decorator returns FAILURE, they don't really work for resetting the timer. At the end, I set up a `TryStopAttackCooldown` in the another branch that is likely to be selected when the AttackSequence failed.

**Additional context**
Tree with the alternative solution implemented, see `TryStopAttackCooldown`.
![image](https://github.com/bitbrain/beehave/assets/10520249/dbcd187c-5d60-4555-a4d6-90f9d3faff92)

rxlecky:
I don't think what you requested is possible because the cooldown is implemented as a decorator node so, as you correctly identified, it doesn't get the `before_run` and `interrupt` callbacks in the same way that action nodes do. You could implement a timer action node to act as your cooldown to achieve the resetting behaviour. Here's a sample implementation that should do the trick:

```gdscript
extends ActionLeaf
class_name WaitAction

@export var wait_time: float = 0.0

@onready var cache_key = 'wait_%s' % self.get_instance_id()

func before_run(actor: Node, blackboard: Blackboard) -> void:
	blackboard.set_value(cache_key, wait_time, str(actor.get_instance_id()))


func tick(actor: Node, blackboard: Blackboard) -> int:
	var remaining_time = blackboard.get_value(cache_key, 0.0, str(actor.get_instance_id()))
	if remaining_time > 0.0:
		remaining_time -= get_physics_process_delta_time()
[...]

PR #392, opened 2025-04-19
Reset decorators correctly

bitbrain:
Related to #340 but not yet fixing it, as propagating interrupts on branch changes does not yet work correctly.


bitbrain:
As stated in #340 we need to interrupt also non-running children to ensure they can reset accordingly. This creates a problem in some cases where the expectation is that interrupt should only be called on running children. I'll see how other famous behaviour tree solutions solve this (e.g. Unreal Engine)

PR #395, opened 2025-04-25
Avoid shared blackboard dictionary by duplicating

### LINKED ISSUE #390: The set_value function of the blackboard is not working as intended.
GODOT: 4.4.1 stable
Beehave: 2.8.0

I implemented a simple patrol AI using Beehave. The patrol AI moves to a target location, and when it reaches a certain distance from the target, it finds a new position and moves again.

When I place this AIPlayer in the scene, it works as intended. However, if I duplicate the AIPlayer and place multiple instances, they don't work as expected.

For example, if there are two players, AIPlayer1 and AIPlayer2, when AIPlayer2 is moving toward a target position, and AIPlayer1's ActionLeaf for finding a new position is executed, it sets the new position in AIPlayer1's blackboard. However, AIPlayer2 ends up heading toward the same new position even though its position-finding ActionLeaf hasn't run.

Why does updating AIPlayer1's target position in its blackboard also change AIPlayer2's blackboard values, causing both players to move toward the same position?

best regards.

Image

[beehave_test_2025-04-17_02-30-52.zip](https://github.com/user-attachments/files/19781762/beehave_test_2025-04-17_02-30-52.zip)

bitbrain:
Closes #390

### DIFF
--- a/addons/beehave/blackboard.gd
+++ b/addons/beehave/blackboard.gd
@@ -7,13 +7,14 @@ const DEFAULT = "default"
 ## multiple nodes of the behavior tree.
 @export var blackboard: Dictionary = {}:
 	set(b):
-		blackboard = b
+		blackboard = b.duplicate()
 		_data[DEFAULT] = blackboard
 
 var _data: Dictionary = {}
 
 
 func _ready():
+	blackboard = blackboard.duplicate()
 	_data[DEFAULT] = blackboard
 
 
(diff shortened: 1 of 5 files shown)


addons/beehave/blackboard.gd
```diff
@@ -7,13 +7,14 @@ const DEFAULT = "default"
 ## multiple nodes of the behavior tree.
 @export var blackboard: Dictionary = {}:
```
bitbrain, line 8: When exporting dictionaries in Godot, they will be shared by default. So we have to manually duplicate them.

PR #410, opened 2025-07-13
Move docstrings above class annotations for better Godot editor integration

csprance:
Repositioned all class-level docstrings to appear before @tool, @icon, and class_name annotations across all behavior tree node files, and added missing docstring to
  RandomizedComposite.

  Why this change was needed

  In Godot 4.x, the "Create Node" dialog window displays class documentation to help developers understand what each node does before adding it to their scene. However, Godot's
  documentation parser only recognizes docstrings that appear before the class annotations, not after them.

  What was changed

  - 23+ node files had their docstrings moved from after the annotations to before them
  - 1 file (randomized_composite.gd) had a missing docstring added
  - 1 file (selector.gd) had single # comments converted to proper ## docstrings

  Example of the change

  Before:
```gdscript
  @tool
  @icon("../../icons/sequence_reactive.svg")
  class_name SequenceReactiveComposite extends Composite

  ## Reactive Sequence nodes will attempt to execute all of its children and report
  ## `SUCCESS` in case all of the children report a `SUCCESS` status code.
```
  After:
```gdscript
  ## Reactive Sequence nodes will attempt to execute all of its children and report
  ## `SUCCESS` in case all of the children report a `SUCCESS` status code.
  @tool
  @icon("../../icons/sequence_reactive.svg")
  class_name SequenceReactiveComposite extends Composite
```
  User benefit

  Now when developers use Godot's "Create Node" dialog (Ctrl+A), they'll see helpful descriptions for each Beehave node type, making it much easier to understand what each node does     
   before adding it to their behavior tree. This significantly improves the developer experience and reduces the need to look up documentation separately.

  Files affected

  All node files in addons/beehave/nodes/ including composites (selectors, sequences, parallel), decorators (cooldown, limiter, inverter, etc.), and leaves (actions, conditions,
  blackboard operations).

### DIFF
--- a/addons/beehave/nodes/beehave_node.gd
+++ b/addons/beehave/nodes/beehave_node.gd
@@ -1,8 +1,7 @@
-@tool
-class_name BeehaveNode extends Node
-
 ## A node in the behavior tree. Every node must return `SUCCESS`, `FAILURE` or
 ## `RUNNING` when ticked.
+@tool
+class_name BeehaveNode extends Node
 
 enum { SUCCESS, FAILURE, RUNNING }
 
--- a/addons/beehave/nodes/beehave_tree.gd
+++ b/addons/beehave/nodes/beehave_tree.gd
@@ -1,8 +1,7 @@
+## Controls the flow of execution of the entire behavior tree.
 @tool
 @icon("../icons/tree.svg")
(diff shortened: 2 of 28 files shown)


bitbrain (approved):
Thank you!

PR #417, opened 2025-10-12
Add leaves in C++

LenNerd42:
## Description

- Add blackboard utility nodes (`BlackboardCompare`, `BlackboardErase`, `BlackboardHas`, `BlackboardSet`)
- Add configuration warnings for all leaves


bitbrain:
@LenNerd42 we also should add docs for these to the wiki in `docs`

PR #419, opened 2025-11-20
Add tree tick_mode

### LINKED ISSUE #389: Behaviour Ticker
For larger scenes, there might be dozens or hundreds of behaviour trees that are ticked every single frame. However, especially for behaviours off-screen this can become a CPU performance bottle-neck. Games like [Kingdom Come Deliverance 2](https://kingdomcomerpg.com) implement a system where NPCs are "ticked" at different rates, depending on if they are nearby or further away.

A system like this for Beehave could be possible with the introduction of a new node, something like `BeehaveSpatialTicker` which would automatically discover all behaviour trees in the scene and manage their ticking (if ticking is set to `MANUAL` only). In case behaviour trees are not set to `MANUAL` it will ignore these and exempt them. Optionally, having a new process mode for trees specifically could also be an option (e.g. `IDLE_MANAGED` and `PHYSICS_MANAGED`), given we might want to tick them in physics or idle thread but that could also be configured on the ticker itself.

From an implementation perspective, this could be solved in various ways:
- viewport checks
- quadtree usage to have gradual `tick()` frequency

Flynsarmy:
## Description

This PR adds a `tick_mode` option to `BeehaveTree` with the following options:
- `BOTH` - ticks on `_process` and `_physics_process`
- `PROCESS` - ticks on `_process`
- `PHYSICS` - ticks on `_physics_process`
- `MANUAL` - doesn't tick at all. Call `$YourTree.tick()` to tick the tree.

This PR is backwards compatible as it defaults to `BOTH` which is the existing functionality. Existing trees will continue working exactly as they always have.

There are a couple of reasons for this change:
- I want to be able to tick some trees at my own custom rate. The `tick_rate` option kind of accomplishes this but setting to `MANUAL` offers the ultimate in flexibility and let's the developer decide exactly how and when the tree ticks
- The current default of `BOTH` is kind of dangerous and your unit tests don't cover this. `_process` runs an unknown number of times per second so `tick_rate` isn't reliable.

Fixes #389

### DIFF
--- a/addons/beehave/nodes/beehave_tree.gd
+++ b/addons/beehave/nodes/beehave_tree.gd
@@ -6,6 +6,7 @@ class_name BeehaveTree extends Node
 enum { SUCCESS, FAILURE, RUNNING }
 
 enum ProcessThread { IDLE, PHYSICS, MANUAL }
+enum TickMode { BOTH, PROCESS, PHYSICS, MANUAL }
 
 signal tree_enabled
 signal tree_disabled
@@ -24,6 +25,10 @@ signal tree_disabled
 
 	get:
 		return enabled
+
+## How automatic ticking of the tree should be handled. The default is to
+## tick on both _process and _physics_process.
+@export var tick_mode: TickMode = TickMode.BOTH
 
 ## How often the tree should tick, in frames. The default value of 1 means
 ## tick() runs every frame.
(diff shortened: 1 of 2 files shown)


joeynelson42:
@Flynsarmy hey, curious if you had any thoughts on your PR. I'm looking for a way to tick the tree manually, did this work as you had hoped?

PR #433, opened 2026-04-05
fixed lib path on linux and added linux specific gitignores

Clicky-coded:
when compiling on linux , lib prefix is added to the library file name so i changed the GDextension file to link properly.
I also added things to .gitignore for linux compilation


bitbrain:
Thanks!

PR #440, opened 2026-07-06
upgrade deps to Godot 4.7

bitbrain:
Godot 4.7 changed some of the GDScript parsing and we had some script bugs that previously did not throw. Those are now addressed.

### DIFF
--- a/addons/beehave/nodes/composites/simple_parallel.gd
+++ b/addons/beehave/nodes/composites/simple_parallel.gd
@@ -33,7 +33,7 @@ func _get_configuration_warnings() -> PackedStringArray:
 	return warnings
 
 
-func tick(actor, blackboard: Blackboard):
+func tick(actor: Node, blackboard: Blackboard) -> int:
 	for c in get_children():
 		var node_index: int = c.get_index()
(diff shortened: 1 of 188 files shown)


addons/beehave/nodes/decorators/repeater.gd
```diff
@@ -10,7 +10,7 @@ class_name RepeaterDecorator extends Decorator
 var current_count: int = 0
 
 
-func before_run(actor: Node, blackboard: Blackboard):
+func before_run(actor: Node, blackboard: Blackboard) -> void:
```
bitbrain, line 13: we should make return value explicit.

addons/beehave/nodes/composites/simple_parallel.gd
```diff
@@ -33,7 +33,7 @@ func _get_configuration_warnings() -> PackedStringArray:
 	return warnings
 
 
-func tick(actor, blackboard: Blackboard):
+func tick(actor: Node, blackboard: Blackboard) -> int:
```
bitbrain, line 36: we should make return value explicit.

PR #443, opened 2026-07-24
Fix singleton conflict caused in godot 4.7.1

### LINKED ISSUE #442: Beehave Singleton Registration possibly break plugins like Phantom Camera
**Godot version: 4.7.1**
**OS: windows 11**

**Describe the bug**
When using Phantom Camera alongside the `Beehave` plugin, Godot throws the error: `Failed to retrieve non-existent singleton 'PhantomCameraManager'` and other scripts are reporting null reference.

The PhantomCameraManager Node is a tool script that is referenced in the editor scene tree,and the scene camera nodes will locate the manager node from `_enter_tree`.

It could be related to the latest Godot 4.7.1 autoload guard fix.
The `Beehave` is working fine but is somehow causing the Godot Editor to kill the `PhantomCameraManager` (the registerd node is lost)
This is the manager node and it would register itself when entering tree.
```gdscript

@tool
extends Node

func _enter_tree() -> void:
	if not Engine.has_singleton(_CONSTANTS.PCAM_MANAGER_NODE_NAME):
		Engine.register_singleton(_CONSTANTS.PCAM_MANAGER_NODE_NAME, self)
	Engine.physics_jitter_fix = 0


func _exit_tree() -> void:
	if Engine.has_singleton(_CONSTANTS.PCAM_MANAGER_NODE_NAME):
		Engine.unregister_singleton(_CONSTANTS.PCAM_MANAGER_NODE_NAME)
```
**To Reproduce**

1. Install both plugins and enable phantom camera first, then enable beehave
2. restart the editor

**Investigation**
I ran several tests and tried to find the cause. It's possibly a conflict in the Godot plugin lifecycle initialization between the two plugins.

1. **Blang Plugin Test:** Created an empty plugin to test Godot's `register_singleton` / `add_autoload_singleton` API works ok.
2. **Single Plugin Test:**  Phantom Camera alone does not produce the error. The bug only manifests when the Beehave plugin is also enabled.
3. **Possible Lifecycle Conflict:** I found that Beehave is registering its singletons inside the `_init()` method:

   ```gdscript
   func _init():
       name = "BeehavePlugin"
       add_autoload_singleton("BeehaveGlobalMetrics", "metrics/beehave_global_metrics.gd")
       add_autoload_singleton("BeehaveGlobalDebugger", "debug/global_debugger.gd")
       
       # Add project settings...
   ```

4. **Workaround Fix:**I modified Beehave's code by moving the singleton registration to `_enable_plugin()`. This does fix the issue on phantom camera plugin.

    ```gdscript
    func _enable_plugin() -> void:
        add_autoload_singleton(METRIC_NAME, "metrics/beehave_global_metrics.gd")
        add_autoload_singleton(DEBUGGER_NAME, "debug/global_debugger.gd")
    ```

5. I also commented out the utils/utils.gd method that searches for the plugin, replacing it with EditorInterface singleton

    ```gdscript
    var border_size := Vector2(4, 4) * EditorInterface.get_editor_scale()
    var editor_main_screen = EditorInterface.get_editor_main_screen()
    ```

**Additional context**
Im not sure if i should report this issue here, since it feels like a godot problem, but just in case others also encountered it and needed the workaround

RawsharkTang:
## Description

Pr for  [Issue442](https://github.com/bitbrain/beehave/issues/442)

## Addressed issues

- modified the plugin initialization process to prevent breaking the godot in-editor singleton 

Closes #442

### DIFF
--- a/addons/beehave/plugin.gd
+++ b/addons/beehave/plugin.gd
@@ -26,7 +26,6 @@ func _enable_plugin() -> void:
 		ProjectSettings.save()
 
 	print("Beehave initialized!")
-	pass
 
 
 func _disable_plugin() -> void:
--- a/addons/beehave/plugin.gd
+++ b/addons/beehave/plugin.gd
@@ -7,24 +7,35 @@ var frames: RefCounted
 var _custom_types: Array[StringName] = []
 
 
-func _init():
-	name = "BeehavePlugin"
+func _enable_plugin() -> void:
 	add_autoload_singleton("BeehaveGlobalMetrics", "metrics/beehave_global_metrics.gd")
 	add_autoload_singleton("BeehaveGlobalDebugger", "debug/global_debugger.gd")
-	
+
 	# Add project settings
 	if not ProjectSettings.has_setting("beehave/debugger/start_detached"):
 		ProjectSettings.set_setting("beehave/debugger/start_detached", false)
 		ProjectSettings.set_initial_value("beehave/debugger/start_detached", false)
-		ProjectSettings.add_property_info({
-			"name": "beehave/debugger/start_detached",
-			"type": TYPE_BOOL,
-			"hint": PROPERTY_HINT_NONE,
-			"hint_string": "If enabled, the debugger will start in a separate window"
-		})
+		ProjectSettings.add_property_info(
+			{
+				"name": "beehave/debugger/start_detached",
+				"type": TYPE_BOOL,
+				"hint": PROPERTY_HINT_NONE,
+				"hint_string": "If enabled, the debugger will start in a separate window"
+			}
+		)
 		ProjectSettings.save()
-	
+
 	print("Beehave initialized!")
+	pass
+
+
+func _disable_plugin() -> void:
+	remove_autoload_singleton("BeehaveGlobalMetrics")
+	remove_autoload_singleton("BeehaveGlobalDebugger")
+
+
+func _init():
+	name = "BeehavePlugin"
 
 
 func _enter_tree() -> void:


addons/beehave/debug/debugger.gd
```diff
@@ -89,7 +85,7 @@ func _on_make_floating() -> void:
 	floating_window.position = editor_main_screen.global_position
 	floating_window.transient = true
 	floating_window.close_requested.connect(_on_window_close_requested)
-	editor_interface.get_base_control().add_child(floating_window)
+	EditorInterface.get_base_control().add_child(floating_window)
```
bitbrain: The problem here is this will break Beehave for anything below Godot 4.2, which is why ideally we should not change this unless it is causing us issues. For the sake of "clean code" we cannot do it, was we'd effectively make Beehave incompatible with anything below Godot 4.2

addons/beehave/debug/debugger_tab.gd
```diff
@@ -14,14 +14,15 @@ var session: EditorDebuggerSession
...
 var active_tree_id = -1
-
 # UI nodes
 var container: HSplitContainer
 var item_list: ItemList
 var graph_container: HSplitContainer
 var graph
 var blackboard_vbox: VBoxContainer
 var message: Label
+static var frame: RefCounted
```
bitbrain: `static` was introduced somewhat during Godot `4.x` development and we should avoid using it, as it makes this addon incompatible with Godot versions where `static` is unsupported.

addons/beehave/debug/debugger_tab.gd
```diff
@@ -2,10 +2,10 @@
 class_name BeehaveDebuggerTab
 extends PanelContainer
 
-const Utils      = preload("res://addons/beehave/utils/utils.gd")
-const TREE_ICON  = preload("../icons/tree.svg")
-const OldGraph   = preload("old_graph_edit.gd")
-const NewGraph   = preload("new_graph_edit.gd")
+# const Utils = preload("res://addons/beehave/utils/utils.gd")
```
bitbrain: make sure to not include any commented out code.

addons/beehave/debug/debugger_tab.gd
```diff
@@ -172,9 +171,10 @@ func _select_item_by_id(id_str: String) -> void:
 		if item_list.get_item_metadata(i) == id_str:
 			item_list.select(i)
 			break
-	get_tree().create_timer(0.2).timeout.connect(func():
-		if item_list.is_inside_tree() and not item_list.get_selected_items().is_empty():
-			_on_item_selected(item_list.get_selected_items()[0])
+	get_tree().create_timer(0.2).timeout.connect(
+		func():
```
bitbrain: @RawsharkTang out of curiosity: what formatter are you using? A lot of files have changed here and we should avoid including unwanted reformatting changes.

We could perhaps agree to adopt the official GDQuest formatter for this repo, but that would need to happen as part of a separate PR/discussion.

addons/beehave/plugin.cfg
```diff
@@ -3,5 +3,5 @@
 name="Beehave"
 description="🐝 Behavior Tree addon for Godot Engine"
 author="bitbrain"
-version="2.9.3-dev"
```
bitbrain, line 6: do not change this

bitbrain:
Thank you for your contribution @RawsharkTang. Unfortunately, this pull request contains a lot more than just the singleton conflict changes (i.e. formatting fixes and cleanups). Ideally, those need to be separated out into separate pull requests, as otherwise it is very difficult for me (in case something breaks or bugs are introduced) what PR exactly caused the change.

Also be mindful that with Beehave, there is a strict compatibility approach where all Godot 4.x versions should be supported. Using things like `static` unfortunately breaks older versions of Godot 4 and the benefit currently does not overwhelm the lockout. Unless we can justify that using something like `static` or `EditorInterface` will solve real bugs for us of course. In that case, we can discuss it :) that being said, I realise this might not be clear to first-time contributors so I will update the contributor guidelines accordingly.

RawsharkTang:
Thanks for the instructions!!! I really dont know much as first time contributing.

addons/beehave/debug/debugger_tab.gd
RawsharkTang: Im using gdtoolkit and gdscript formatter to format my code.

RawsharkTang:
Im closing this pr and i will open up a new one with only the conflict fix

RawsharkTang:
I just forced push and found this pr's old change is lost ahhhh
Update: It seems I can just reuse this pr and dont have to reopen one.
I cleaned up my commit and only preserve the conflict fix

addons/beehave/plugin.gd
```diff
@@ -7,24 +7,35 @@ var frames: RefCounted
...
+				"type": TYPE_BOOL,
+				"hint": PROPERTY_HINT_NONE,
+				"hint_string": "If enabled, the debugger will start in a separate window"
+			}
+		)
 		ProjectSettings.save()
-	
+
 	print("Beehave initialized!")
+	pass
```
bitbrain: @RawsharkTang minor: remove the `pass`
```suggestion
```

bitbrain (approved):

